PR #10, opened 2019-01-30
upload finished and prettier run over code

danielhalupka:


danielhalupka:
@EdytaSki Those are compiled lines of code into ES5, don't pay attention to compiled code please.
We need to bundle code with babel before publishing to be able to use it as node_module import in react, these needs to look as such. It gets minified in the final product after all :)

PR #22, opened 2019-03-15
 enh(ui): better display of errors on file upload

kduret:
Improving the display of errors and getting around bind call using experimental plublic class fields syntax


jdelpierre:
a small point to note: Your first commit named "tmp" is squash if your PR is approuved?

PR #53, opened 2019-04-21
 enh(ui): improve license display in card footer

kduret:
+ remove license button from extension detailed popin

### DIFF
--- a/src/ExtensionsHolder/index.js
+++ b/src/ExtensionsHolder/index.js
@@ -18,7 +18,8 @@ class ExtensionsHolder extends React.Component {
     return description.replace(/^centreon\s+(\w+)/i, (_, $1) => $1);
   }
 
-  checkLicense = licenseInfo => {
+  // get CardItem props to display license information in footer
+  getPropsFromLicense = licenseInfo => {
     let licenseProps = {};
 
     if (licenseInfo && licenseInfo.required) {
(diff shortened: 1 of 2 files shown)


src/ExtensionsHolder/index.js
```diff
@@ -18,10 +18,29 @@ class ExtensionsHolder extends React.Component {
...
+    if (licenseInfo && licenseInfo.required) {
+      if (!licenseInfo.expiration_date) {
+        licenseProps.itemFooterColor = "red";
+        licenseProps.itemFooterLabel = "Your license is not valid";
+      } else {
+        // @todo move this logic to centreon. Furthermore, it will facilitate translation
+        // @todo use moment to convert date in the proper format (locale and timezone from user)
+        const expirationDate = (new Date(licenseInfo.expiration_date)).toISOString().slice(0,10);
+        licenseProps.itemFooterColor = "green";
+        licenseProps.itemFooterLabel = `Expiration date : ${expirationDate}`;
```
quanghungb: I would prefer to have the same "information" for the 2 cases (in order to be consistent) :
- Your license is not valid
and
- Your license is valid until xxxx

quanghungb: As discussed, it would be more consistent by talking both about license or module
- License is not valid
- License is valid until yyyyy-mm-dd

src/ExtensionsHolder/index.js
```diff
@@ -18,10 +18,29 @@ class ExtensionsHolder extends React.Component {
...
+    if (licenseInfo && licenseInfo.required) {
+      if (!licenseInfo.expiration_date) {
+        licenseProps.itemFooterColor = "red";
+        licenseProps.itemFooterLabel = "License is not valid";
+      } else {
+        // @todo move this logic to centreon. Furthermore, it will facilitate translation
+        // @todo use moment to convert date in the proper format (locale and timezone from user)
+        const expirationDate = (new Date(licenseInfo.expiration_date)).toISOString().slice(0,10);
+        licenseProps.itemFooterColor = "green";
+        licenseProps.itemFooterLabel = `License is valid until : ${expirationDate}`;
```
jdelpierre: I think it would be good to check that the format is correct and if not to display the two points with the date « expirationDate » no?

PR #65, opened 2019-06-04
Mon 3531 icons svg hd

EdytaSki:


EdytaSki:
Will create another one

PR #69, opened 2019-06-13
MON-3531 #comment new commit with cleaned files

EdytaSki:


src/index.js
```diff
@@ -46,6 +42,8 @@ export { default as MessageError} from './Message/MessageError';
 export { default as MessageInfo} from './Message/MessageInfo';
 export { default as MessageStatus } from "./Message/MessageStatus";
 export { default as Navigation } from "./Navigation";
+export { default as NativeSelects } from "./FormMui";
+export { default as PageHdModifiers } from "./PageHdModifiers";
```
kduret: Did you test it ?
Cause I guess it will break cause those components are not on master branch...

EdytaSki: yes you are right - I will delete this.
I tested in my local with the hd-modifiers dev so that's why I didn't notice. Thanx.

bdauria:
@EdytaSki What's the status here?

bdauria:
@EdytaSki up

PR #80, opened 2019-06-24
Bam list add edit delete

danielhalupka:


src/InputField/InputFieldMultiSelectNew/index.js
```diff
@@ -0,0 +1,43 @@
+import React, { Component } from "react";
+import classnames from "classnames";
+import styles from "./input-field-multi-select.scss";
+
+class InputFieldMultiSelect extends Component {
```
bdauria: I think we can render stateless Components as function, since we are planning on using functions only after migrating to React v16.8.

@danielhalupka you mentioned that it was easier to use Classes with material-ui for some reason but I can't remember why, can you remind me?

src/MaterialComponents/Icons/IconEdit/index.js
```diff
@@ -0,0 +1,30 @@
...
+    position: 'absolute',
+    right: 3,
+  },
+}));
+
+function IconEdit({customStyle, onClick}) {
+  const classes = useStyles();
+
+  return (
+    <React.Fragment>
```
bdauria: I think we can get rid of the Fragment here since we only have one element

src/MaterialComponents/Switch/index.js
```diff
@@ -0,0 +1,36 @@
...
+    '&$checked + $track': {
+      backgroundColor: '#0072CE',
+      opacity: '.4'
+    },
+  },
+  checked: {},
+  track: {},
+});
+
+class CustomSwitch extends Component {
```
bdauria: Same remark as above regarding functional vs class, if we figure out that classes are not needed.

src/MultiSelectHolder/index.js
```diff
@@ -0,0 +1,47 @@
...
+    return (
+      <div
+        className={classnames(
+          styles["multi-select-holder"],
+          isEmpty ? styles["multi-select-holder-empty"] : ""
+        )}
+        onMouseEnter={this.toggleHover.bind(this)}
+        onMouseLeave={this.toggleHover.bind(this)}
+      >
+        {isEmpty && (
```
bdauria: Can we factorise the two span elements within one isEmpty check to eliminate some duplication?

src/MultiSelectHolder/index.js
```diff
@@ -0,0 +1,47 @@
...
+  render() {
+    const { isEmpty, multiSelectCount, multiSelectLabel, children } = this.props;
+    const { isHovered } = this.state;
+    return (
+      <div
+        className={classnames(
+          styles["multi-select-holder"],
+          isEmpty ? styles["multi-select-holder-empty"] : ""
+        )}
+        onMouseEnter={this.toggleHover.bind(this)}
```
bdauria: Do we need to bind to this if we're using an arrow method?

src/Panels/index.js
```diff
@@ -0,0 +1,109 @@
...
+            styles[panelItemActive ? "panels-second-active" : ""]
+          )}
+        >
+          <div className={classnames(styles["panels-dialog"])}>
+            <div className={classnames(styles["panels-inner"])}>
+              <div className={classnames(styles["panels-header"])}>
+                <IconAttach />
+                <IconPowerSettings active={true} />
+                <PanelHeaderTitle label="Africa office availability" />
+                <IconCloseNew onClick={this.togglePanel.bind(this)} />
```
bdauria: Same remark as above concerning binding & arrow methods

src/Panels/index.js
```diff
@@ -0,0 +1,109 @@
...
+                >
+                  <div className={classnames(styles2["panel-item-inner"])}>
+                    <Accordion />
+                  </div>
+                  <span
+                    className={classnames(
+                      styles["panels-arrow"],
+                      panelItemActive ? styles["panels-arrow-right"] : ""
+                    )}
+                    onClick={this.toggleSecondPanel.bind(this)}
```
bdauria: same remark concerning bind & arrow method

src/Submenu/SubmenuHeader/Mocked-Submenu.js
```diff
@@ -0,0 +1,91 @@
...
+        <div
+          className={classnames(
+            styles["submenu-top"],
+            styles[ active ? "submenu-active" : ""]
+          )}
+          style={{
+            width: "200px"
+          }}
+        >
+          <IconHeader iconType="services" iconName="services" onClick={this.toggleSubmenu.bind(this)}>
```
bdauria: Same remark concerning bind & arrow method

src/Submenu/SubmenuHeader/Mocked-Submenu.js
```diff
@@ -0,0 +1,91 @@
...
+            iconColor="gray-dark"
+            iconNumber="5"
+          />
+          <IconNumber iconType="colored" iconColor="green" iconNumber="10" />
+          <IconToggleSubmenu
+            iconPosition="icons-toggle-position-right"
+            iconType="arrow"
+            onClick={this.toggleSubmenu.bind(this)}
+          />
+          <div className={classnames(styles["submenu-toggle"])}>
```
bdauria: I think we can create easily extract a Component since there's no real logic appart from displaying. How about Submenu?

src/MaterialComponents/Icons/IconEdit/index.js
NikolaL87: Yes. Unnecessary Fragment tag is removed now

src/MultiSelectHolder/index.js
NikolaL87: Yes. The two span elements are in one condition now.

src/InputField/InputFieldMultiSelectNew/index.js
danielhalupka: This particular one is not stateless, will have to have state, for the sake of implementation

src/Submenu/SubmenuHeader/Mocked-Submenu.js
danielhalupka: Don't you see it's Mocked? It's just an example :)

src/Submenu/SubmenuHeader/Mocked-Submenu.js
danielhalupka: Please ignore files that has Mocked prefix, they are not used in the project.

src/Panels/index.js
danielhalupka: This PR concerns listing and Panels are in the state of implementation after this is merged, so don't worry, you'll get totally new PR for the whole part of Panels and form that it holds. Please ignore it for now

src/Panels/index.js
danielhalupka: Same as bellow

src/MultiSelectHolder/index.js
danielhalupka: Part of panel implementation, this PR is made to be able to validate listing, so, this is still not implemented, it's just an UI example that will be redone

src/MaterialComponents/Switch/index.js
danielhalupka: Right now, it seems to be stateless, not sure it will stay stateless when we implement panels.

bdauria: We won't have this issue anymore with https://github.com/centreon/centreon-react-components/pull/81 ;)

PR #88, opened 2019-07-04
fix(ui): Small fix for bam listing actions and codebase formated with prettier

danielhalupka:


src/Pages/BAMListing.js
```diff
@@ -119,6 +119,7 @@ class BAMListingPage extends Component {
     this.setState({
       duplicateActive: !duplicateActive
     });
+    aster;
```
bdauria: Looks like we can delete that

danielhalupka: done

PR #95, opened 2019-07-09
fix(ui): Fix and optimize BAM listing pages

NikolaL87:


NikolaL87:
> @NikolaL87 Can you change the PR name to something more meaningful?

> @bdauria Yes. Is this good enough now?

src/Table/TableCustom/index.js
```diff
@@ -157,6 +161,11 @@ class TableCustom extends Component {
                       key={row.id}
                       selected={isItemSelected}
                       onMouseEnter={this.rowHovered.bind(this, row.id, true)}
+                      {...(enabledColumn && {
```
kduret: could be splitted in a separate function

src/Table/TableCustom/index.js
```diff
@@ -35,6 +35,9 @@ const styles = (theme) => ({
   tableWrapper: {
     overflowX: 'auto',
   },
+  rowDisabled: {
+    backgroundColor: 'rgba(0, 0, 0, 0.07) !important',
```
kduret: why do you use !important ?

PR #96, opened 2019-07-09
fix(test): Adjust acceptance test related markup

victorvassilev:


src/Button/ButtonCustom/index.js
```diff
@@ -24,15 +24,15 @@ const useStyles = makeStyles((theme) => ({
   },
 }));
 
-function ButtonCustom({ label, onClick }) {
+function ButtonCustom({ label, onClick, id }) {
```
kduret: I think it should not be a mandatory prop.
Use ...rest instead

kduret:
```suggestion
function ButtonCustom({ label, onClick, ...rest }) {
```

src/Button/ButtonCustom/index.js
```diff
@@ -24,15 +24,15 @@ const useStyles = makeStyles((theme) => ({
...
   const classes = useStyles();
-
   return (
     <Button
       variant="contained"
       color="secondary"
       className={classes.button}
       onClick={onClick}
+      id={id}
```
kduret:
```suggestion
      {...rest}
```

bdauria:
@victorvassilev any update on this?

PR #97, opened 2019-07-10
fix(ui): Fix and optimize BAM listing page

NikolaL87:


src/Breadcrumb/Link.js
```diff
@@ -11,9 +10,9 @@ const BreadcrumbLink = ({ index, count, breadcrumb, classes }) => {
       {breadcrumb.label}
     </Link>
   ) : (
-    <Typography className={classes.item} color="textPrimary">
+    <Link className={classes.item} href={breadcrumb.link} color="textPrimary">
```
kduret:
```suggestion
    <Link className={classes.item} to={breadcrumb.link} color="textPrimary">
```

src/Table/TableCustom/index.js
```diff
@@ -35,6 +35,9 @@ const styles = (theme) => ({
   tableWrapper: {
     overflowX: 'auto',
   },
+  rowDisabled: {
+    backgroundColor: 'rgba(0, 0, 0, 0.07) !important',
```
kduret: why do you use !important ?

src/Table/TableCustom/index.js
```diff
@@ -157,6 +161,11 @@ class TableCustom extends Component {
                       key={row.id}
                       selected={isItemSelected}
                       onMouseEnter={this.rowHovered.bind(this, row.id, true)}
+                      {...(enabledColumn && {
```
kduret: could be splitted in a separate function

src/Table/TableCustom/index.js
```diff
@@ -309,7 +342,7 @@ class TableCustom extends Component {
             </Table>
           </div>
           <StyledPagination
-            rowsPerPageOptions={[5, 10, 25, 50, 100]}
+            rowsPerPageOptions={[10, 20, 30, 40, 50, 60, 70, 80, 90, 100]}
```
kduret: Maybe too much values ?
IMO it should be nice that the max value corresponds to the total count of items

src/Pages/BAMListing.js
```diff
@@ -286,6 +286,7 @@ class BAMListingPage extends Component {
             onEnable={onEnable}
             onDisable={onDisable}
             selected={currentlySelected}
+            enabledColumn="activate"
```
bdauria: What exactly does ```enabledColumn``` mean? it looks like it's used as a boolean but a string is given

src/Breadcrumb/Link.js
NikolaL87: If we put like this than the "href" will be changed with "to" only ( without the path ). It doesn't work like classical Link at the React.

src/Table/TableCustom/index.js
NikolaL87: There is a CSS pseudo selector for odd rows. The important is here because pseudo selector is stronger than the regular one, and there are some cases that style won't apply on the table row.

src/Table/TableCustom/index.js
NikolaL87: We can discuss about this. There is a task that was created to insert the values increasing by 10, but we can change this for sure.

src/Pages/BAMListing.js
NikolaL87: It's a prop that is passed to the component that can do some logic if the entity is activated, or add class disabled to the row if it's not.

src/Table/TableCustom/index.js
kduret: but this could be managed directly in StyleTableRow, without using !important
"styles" can be a function which takes as argument props of the component

src/Table/TableCustom/index.js
bdauria: I feel like we could benefit from generating the values through a function so we can change the parameters easier in the future.

What about something like:

```
function generateInterval({start, increment, amount}) {
  return Array(amount).fill().map((_, i) => i * increment + start);
}
```

src/Table/TableCustom/index.js
NikolaL87: Even if we put in the StyledTableRow there are some cases that pseudo selector will be stronger than the regular one. It's CSS related only. I can put on the same place without important but will not cover all cases.

src/Table/TableCustom/index.js
NikolaL87: As I talked with Danijel this will be refactored in the future for sure.

bdauria: @NikolaL87 That's understandable, but maybe having a separate function would help the refactoring afterwards?

src/Table/TableCustom/index.js
NikolaL87: Yes. We can do like this.

src/Pages/BAMListing.js
danielhalupka: So that you get it be dynamic,  you can specify the column by which the table will mark disabled/enabled rows, since this table will be used in places which do not have activate column in their tableData.

src/Table/TableCustom/index.js
danielhalupka: No need to refactor this, in the end, this param will come from default values on the BE, when endpoint is ready.

src/Breadcrumb/Link.js
danielhalupka: Since breadcrumbs are not only linked to inside of react route, we use href to link it to pages that are not react. Leave it as is

src/Table/TableCustom/index.js
NikolaL87: Done.

src/Table/TableCustom/index.js
NikolaL87: Important was removed. Done.

src/Table/TableCustom/index.js
```diff
@@ -161,11 +176,13 @@ class TableCustom extends Component {
                       key={row.id}
                       selected={isItemSelected}
                       onMouseEnter={this.rowHovered.bind(this, row.id, true)}
-                      {...(enabledColumn && {
-                        className: !row[enabledColumn]
-                          ? classes.rowDisabled
-                          : '',
-                      })}
+                      {...this.addConditionalRowBackground(
```
bdauria: I wouldn't go as far as making everything as parameters. Most of the stuff is already available, either through props or plain string. I think the only parameter should be row in that case.

danielhalupka: Then we lose dynamics and we specify stuff directly, which sould not be the cas, this way you can reuse the function, there is no point in creating static function just so that its a function. Think dynamic

bdauria: you are right, making things dynamic usually make sense, but in this particular example, it comes at the expense of readability, which was the idea behind @kduret 's request. Plus we still have the dynamic aspect of getting some parameters through the props. 

@NikolaL87 @danielhalupka any thoughts?

PR #108, opened 2019-07-29
feat(ui): Add error dialog component

bdauria:


src/MaterialComponents/Dialog/Error/index.test.jsx
```diff
@@ -0,0 +1,21 @@
...
+      <ErrorDialog
+        open
+        title="Error"
+        text="Something unexpected happened..."
+        confirmLabel="Close"
+        onClose={() => {}}
+      />,
+    );
+
+    expect(container.firstChild).toMatchSnapshot();
```
kduret: Avoid to get an empty snapshot :
```suggestion
    expect(container).toMatchSnapshot();
```

package.json
```diff
@@ -89,7 +89,6 @@
     "eslint-config-airbnb": "^17.1.0",
     "eslint-config-prettier": "^4.3.0",
     "eslint-plugin-import": "^2.17.3",
-    "eslint-plugin-jest": "^22.7.2",
```
kduret: why do you remove this dependency ?

src/MaterialComponents/Dialog/Error/index.test.jsx
bdauria: > Avoid to get an empty snapshot :

It doesn't really help, we just get an empty container instead. I think the problem comes from the mui animation, I'll leave a get around this later on.

package.json
bdauria: I made a fix in eslint-config-centreon. There is a property that can be set in the env object to allow just namespace. see https://github.com/centreon/eslint-config-centreon/commit/d196ef77510f220affed3b82e12d160a9974d87a

PR #109, opened 2019-07-29
fix(ui): Fix PromptDialog Apply label

bdauria:


src/Dialog/Prompt/index.js
```diff
@@ -64,7 +64,7 @@ class PromptDialog extends Component {
               onYesClicked(prompt);
             }}
           >
-            Appy
+            Apply
```
bdauria: We have to make the label customizable through props to be able to translate it within the modules.

PR #117, opened 2019-09-09
enh(ui): Enhance Loading indicator in TableCustom

bdauria:


src/Table/TableCustom/index.js
```diff
@@ -661,17 +675,14 @@ class TableCustom extends Component {
...
-                    >
-                      <CircularProgress className={classes.loadingIndicator} />
-                    </TableCellCustom>
-                  </StyledTableRow>
+                  <TableRow tabIndex={-1} className={classes.loadingRow}>
+                    <CircularProgress className={classes.loadingIndicator} />
+                  </TableRow>
                 )}
-                {tableData.length < 1 && !loading ? (
+                {loading && tableData.length < 2 && (
```
bdauria: We need a Row that's big enough if we have less than two entries in order to display the Loading indicator properly

PR #121, opened 2019-09-16
fix(ui): Fix Save Button on BA page and adjust styles to match BV

bdauria:


src/MultiSelectPanel/index.js
```diff
@@ -101,22 +103,23 @@ class MultiselectPanel extends Component {
               <CustomColumn customColumn="md-6">
                 <FormControlLabel
                   labelPlacement="top"
-                  control={
-                    <MaterialSwitch
+                  control={(
+<MaterialSwitch
```
kduret: indentation seems bad

src/MultiSelectPanel/index.js
```diff
@@ -101,22 +103,23 @@ class MultiselectPanel extends Component {
...
                       onChange={onlySelectedChange}
                     />
-                  }
-                  label={
-                    <Typography style={{
-                      fontSize: '13px',
-                    }}
+)}
+                  label={(
+<Typography
```
kduret: indentation seems bad

PR #127, opened 2019-09-20
Fix/707 701 704 699

danielhalupka:


src/MaterialComponents/Icons/IconAttach/index.jsx
```diff
@@ -1,35 +1,78 @@
+/* eslint-disable jsx-a11y/no-static-element-interactions */
```
bdauria: Can we fix these issues instead of manually disabling through comments?

danielhalupka: I'm working on that, but it the other issue.

PR #140, opened 2019-10-15
enh(ui): TableCustom styling issues and refactor

bdauria:


src/Table/TableCustom/index.js
```diff
@@ -228,38 +245,51 @@ class TableCustom extends Component {
 
     const emptyRows = limit - Math.min(limit, totalRows - currentPage * limit);
 
+    const tableMaxHeight = () => {
+      const { tableHeight, tableTopOffset } = this.state;
+
+      return `calc(100vh - ${tableHeight}px - ${tableTopOffset}px - 5px)`;
```
bdauria: There's surely a better way, but that's the only one I found so far.

At least we're a bit more precise now (instead of guessing the offset)

src/Table/TableCustom/EnhancedTableHead.js
```diff
@@ -1,38 +1,103 @@
...
+import TableCell from '@material-ui/core/TableCell';
+import { withStyles, createStyles } from '@material-ui/core/styles';
 import StyledTableSortLabel from './StyledTableSortLabel';
 import StyledCheckbox from './StyledCheckbox';
 import TABLE_COLUMN_TYPES from '../ColumnTypes';
 
-class EnhancedTableHead extends Component {
-  createSortHandler = (property) => (event) => {
-    const { onRequestSort } = this.props;
+const HeaderCell = withStyles(() =>
```
kduret: styled could avoid usage of withstyles and makestyles

src/Table/TableCustom/EnhancedTableHead.js
```diff
@@ -63,85 +120,42 @@ class EnhancedTableHead extends Component {
...
+              padding="none"
+              style={
+                indicatorsEditor
+                  ? {
+                      padding: '3px 4px',
+                    }
+                  : {}
+              }
+            >
+              Unknown
```
kduret: we'll need to handle translation. But I guess we'll move this kind of information in the bam repository...

src/Table/TableCustom/index.js
kduret: IE11 is not compatible with this kind of calculation, but I think it does not break the UI. (I had the same issue on another repo ^^)

src/Table/TableCustom/index.js
```diff
@@ -268,450 +298,432 @@ class TableCustom extends Component {
...
+                ref={this.tableBodyRef}
+                onMouseLeave={this.rowHovered.bind(this, '', false)}
+                style={{
+                  position: 'relative',
+                }}
+              >
+                {tableData.map((row, index) => {
+                  const isItemSelected = isSelected(
+                    indicatorsEditor
+                      ? row
```
kduret: we'll probably need to move it in a separate function later

bdauria: we'lle also refactor out all the indicators related stuff

PR #145, opened 2019-11-08
feat(wizard): add wizard component (with formik)

kduret:
Refs: MON-4169


src/Wizard/StepIcon.jsx
```diff
@@ -0,0 +1,71 @@
...
+  },
+});
+
+function StepIcon({ active, completed, icon }) {
+  const classes = useStepIconStyles();
+
+  return (
+    <div className={classes.root}>
+      {completed ? (
+        <Avatar className={classNames(classes.avatar, classes.avatarCompleted)}>
```
bdauria:
```suggestion
        <Avatar className={${classes.avatar}  ${classes.avatarCompleted}}>
```

bdauria: We don't need classnames, material-ui supports styles composition

src/Wizard/index.jsx
```diff
@@ -0,0 +1,210 @@
...
+    return true;
+  }
+
+  return false;
+}
+
+function cloneElement(element, props) {
+  const forwardedProps = isReactElement(element) ? props : {};
+
+  console.log(isReactElement(element))
```
bdauria: Remove console.logs

src/Dialog/Confirm/index.jsx
```diff
@@ -0,0 +1,50 @@
...
+}) {
+  return (
+    <Dialog
+      open={open}
+      onClose={onClose}
+      onCancel={onCancel}
+      onConfirm={onConfirm}
+      labelTitle={labelTitle}
+      labelCancel={labelCancel}
+      labelConfirm={labelConfirm}
```
bdauria: Can't we transfer the props here instead?

src/Wizard/StepIcon.jsx
```diff
@@ -21,26 +21,26 @@ const useStepIconStyles = makeStyles({
   },
   avatarActive: {
     color: '#ffffff',
-    backgroundColor: '#009fdf',
-    boxShadow: '0 3px 2px 1px rgba(0,0,0,.25)',
+    backgroundColor: theme.palette.primary.main,
```
bdauria: Nice!

src/Dialog/Confirm/index.jsx
kduret: I wanted to avoid warning about spread props.
But indeed, I will transfer props, and we will add a rule in eslint-config-centreon

kduret: fixed (also in Duplicate dialog)

PR #146, opened 2019-11-16
feat(component): add search input component

kduret:
image
focused :
image


src/InputField/SearchInput/index.jsx
```diff
@@ -0,0 +1,87 @@
...
+import SearchIcon from '@material-ui/icons/Search';
+
+const useAdornmentStyles = makeStyles((theme) => ({
+  searchIcon: {
+    color: 'darkgray',
+    '& > input$focused': {
+      color: theme.palette.primary.main,
+    },
+  },
+  focused: {},
```
jdelpierre: Why "focused" is empty, Is necessary to declare it ?

src/InputField/SearchInput/index.jsx
```diff
@@ -0,0 +1,87 @@
...
+    </InputAdornment>
+  );
+}
+
+const useStyles = makeStyles((theme) => {
+  // took it from material-ui source code
+  const light = theme.palette.type === 'light';
+  const bottomLineColor = light
+    ? 'rgba(0, 0, 0, 0.42)'
+    : 'rgba(255, 255, 255, 0.7)';
```
jdelpierre: the "rgba" is not defined in global theme ? 
If we add another field, we will have to redefine these colors?

src/InputField/SearchInput/index.jsx
kduret: it is mandatory to link it to searchIcon property ('& > input$focused)

src/InputField/SearchInput/index.jsx
kduret: I wrote in the comment "took it from material-ui source code" ;)
So, event material-ui does not use it as theme :D

src/InputField/SearchInput/index.jsx
jdelpierre: Ok ;)

src/InputField/SearchInput/index.jsx
jdelpierre: Ah okay sorry I didn't look at the associated commentary properly :(

PR #147, opened 2019-11-29
enh(wizard): forward formik bag in props

kduret:
* forward Formik's bag property in wizard pages
* disable actionbar's next button if form is not valid
* use theme in SearchInput component

### DIFF
--- a/src/Wizard/ActionBar.jsx
+++ b/src/Wizard/ActionBar.jsx
@@ -3,20 +3,20 @@ import PropTypes from 'prop-types';
 import Grid from '@material-ui/core/Grid';
 import Button from '@material-ui/core/Button';
 
-function ActionBar(props) {
-  const {
-    page,
-    isLastPage,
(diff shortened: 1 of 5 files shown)


src/Wizard/ActionBar.jsx
```diff
@@ -3,20 +3,20 @@ import PropTypes from 'prop-types';
...
-    onPrevious,
-    onNext,
-    onFinish,
-    labelCancel,
-    labelPrevious,
-    labelNext,
-    labelFinish,
-  } = props;
+function ActionBar({
+  bag: { isSubmitting, isValid },
```
bdauria: I'm not sure whether bag makes sense for a Props name here. It's a very specific formik term, but I'm afraid it's not very clear in isolation. 
I would leave the form element outside of the Component and pass a disabled Props with the check already done. I understand there will be some duplication but at least the scope will be more clear.

PR #149, opened 2019-12-05
feat(textfield): add textfield component

kduret:
+ introduce clsx, which is lightweight compared to classnames

image


src/TextField/index.jsx
```diff
@@ -0,0 +1,81 @@
...
+) {
+  const classes = useStyles();
+
+  return (
+    <MuiTextField
+      InputProps={{
+        ...InputProps,
+        classes: {
+          ...InputProps.classes,
+          root: clsx(
```
bdauria: Not entirely sure why we need clsx. Mui already supports class composition, right?

src/TextField/index.jsx
```diff
@@ -0,0 +1,81 @@
...
+  },
+  inputHelperText: {
+    marginTop: 4,
+    textAlign: 'right',
+  },
+}));
+
+const TextField = forwardRef(function TextField(
+  { InputProps, InputLabelProps, FormHelperTextProps, ...other },
+  ref,
```
bdauria: For what use case do you need the initial ref?

kduret: one of the use case is to manage focus

src/TextField/index.jsx
kduret: If I instantiate centreon-ui/TextField and I want to override "input" class, I need to use clsx to keep properties of this component (L12).
Do you know an other way to do it ?

kduret: If i override only root, I can probably just use className prop, but in the case of nested class, I need to use this way

PR #151, opened 2019-12-14
fix(ui): Fix Indicators editor, add possibility of default sorting in TableCustom and add aria-labels

bdauria:


src/InputField/InputFieldSelectTableCell/index.jsx
```diff
@@ -19,6 +19,25 @@ class InputFieldSelectCustom extends Component {
     selected: {},
   };
 
+
+  componentWillMount = () => {
+    const { value, options } = this.props;
+    if (options) {
+      for (let i = 0; i < options.length; i += 1) {
```
kduret:
```suggestion
      for (const i = 0; i < options.length; i += 1) {
```

src/MaterialComponents/Snackbar/Error/withErrorSnackbar.jsx
```diff
@@ -16,10 +16,29 @@ const withErrorSnackbar = (Component) => {
...
+    const showErrors = (errors) => {
+      const errorKeys = Object.keys(errors);
+
+      const formattedErrors = errorKeys.reduce(
+        (acc, current) => [...acc, `${current}: ${errors[current]}`],
+        [],
+      );
+  
+      showError(
+        <div style={{ display: 'block' }}>
```
kduret: could you move style in a separate object and use useStyles ?

src/MaterialComponents/Snackbar/Error/withErrorSnackbar.jsx
```diff
@@ -16,10 +16,29 @@ const withErrorSnackbar = (Component) => {
...
+          {formattedErrors.map((err, index) => (
+            <p style={{ margin: 0 }} key={errorKeys[index]}>
```
kduret: could you move style in a separate object and use useStyles ?

src/MaterialComponents/Snackbar/Error/withErrorSnackbar.jsx
bdauria: Actually the block is displayed statically so there is no way to use Hooks.

PR #157, opened 2020-01-07
feat(stories): Add stories iconheader and iconnumber

Thebarda:
This PR adds stories about IconHeader and IconNumber.


src/Icon/IconHeader/index.stories.jsx
```diff
@@ -0,0 +1,137 @@
+/* eslint-disable no-alert */
+/* eslint-disable react/prop-types */
+
+import React from 'react';
+import styles from './stories.scss';
```
bdauria: can we use mui styling instead?

src/Icon/IconHeader/index.stories.jsx
```diff
@@ -0,0 +1,136 @@
...
+  display: 'block',
+  width: '50px',
+};
+
+const alertOnClick = (name) => {
+  alert(`${name} clicked`);
+};
+
+const HeaderBackground = ({ children }) => (
+  <div style={{ backgroundColor: '#232f39' }}>{children}</div>
```
bdauria: It looks like this background color makes a few icon less visible. Can we fix that?

PR #158, opened 2020-01-08
fix(wizard): use last updated values to step validation

kduret:


src/Wizard/index.jsx
```diff
@@ -86,10 +86,12 @@ function Wizard(props) {
     setPage(Math.max(page - 1, 0));
   };
 
-  const validate = () => {
+  const validate = (valuesToValidate) => {
```
kduret:
```suggestion
  const validate = (currentValues) => {
```

src/Wizard/index.jsx
```diff
@@ -86,10 +86,12 @@ function Wizard(props) {
...
     const activePage = React.Children.toArray(children)[page];
 
-    return activePage.props.validate ? activePage.props.validate(values) : {};
+    return activePage.props.validate
+      ? activePage.props.validate(valuesToValidate)
```
kduret:
```suggestion
      ? activePage.props.validate(currentValues)
```

PR #160, opened 2020-01-10
refacto(listing): Rename TableCustom to Listing and remove KPI logic

bdauria:


src/InputField/InputFieldSelectTableCell/index.jsx
```diff
@@ -22,38 +23,40 @@ class InputFieldSelectCustom extends Component {
...
+    if (!options) {
+      return;
+    }
+
+      const valueOption = options.find(({ id }) => id === value);
+
+      if (valueOption) {
+        this.setState({ selected: valueOption})
       }
+
```
Thebarda: Is it possible to create a function that update state for a selected value to deduplicate the code ?

bdauria: Yes, that would be ideal. This Component will be removed soon though, because it doesn't serve any purpose (will be replaced by our unified Input)

PR #163, opened 2020-01-20
enh(theme): Customize error and warning on theme palette

Thebarda:


kduret:
> I think we can live with the default colors of mui, can't we?

we can for the error color.
But I think it does not exist "warning" color in default theme https://material-ui.com/fr/customization/default-theme/#explorer

jdelpierre:
> > I think we can live with the default colors of mui, can't we?
> 
> we can for the error color.
> But I think it does not exist "warning" color in default theme https://material-ui.com/fr/customization/default-theme/#explorer

Hum I think it's declared by base on this page  ? 
https://material-ui.com/fr/customization/palette/

Thebarda:
> @Thebarda what's the status here, do we still need that PR?

I think, we do not neet this PR anymore, I will close it

PR #167, opened 2020-01-24
enh(core): upgrade storybook and remove dependencies

kduret:
- upgrade storybook to 5.3 (configuration update)
- replace emotionjs by material-ui styling solution
- replace classnames by clsx (lightweight)
- remove useless dependencies


Thebarda:
Do we need to use clsx instead of classnames inside Centreon Web, BAM and Auto Disco ?

kduret:
> Do we need to use clsx instead of classnames inside Centreon Web, BAM and Auto Disco ?

I think so cause it's more lightweight than classnames and it's the solution used by material-ui
Furthermore, we don't use extra features provided by classnames

src/storyshots.test.js
```diff
@@ -14,12 +14,19 @@ const beforeScreenshot = () => {
     }, 600),
   );
 };
+const getScreenshotOptions = () => {
```
bdauria: Why is this change necessary?

kduret: Notre necessary 
At the beginning, I Wanted to remove background but finally I let it.
Then i will remove this function

PR #174, opened 2020-02-12
enh(ui): Refactor Listing, cleanup warning messages and rework MultiSelectEntries

bdauria:


src/Listing/Pagination.tsx
```diff
@@ -1,3 +1,5 @@
+import React from 'react';
+
 import TablePagination from '@material-ui/core/TablePagination';
 import { withStyles } from '@material-ui/core/styles';
```
Thebarda: I think we can use named import here

PR #179, opened 2020-02-18
feat(ui): Create Components for unified view and adapt styles

bdauria:


src/Listing/Header.tsx
```diff
@@ -1,21 +1,25 @@
...
+
 import StyledCheckbox from './Checkbox';
-import TABLE_COLUMN_TYPES from './ColumnTypes';
 
-const HeaderCell = withStyles({
+const HeaderCell = withStyles((theme) => ({
   root: {
-    backgroundColor: '#009fdf',
-    color: '#fff',
+    // color: theme.palette.common.black,
```
kduret: remove comments

package.json
```diff
@@ -29,16 +29,16 @@
...
+    "react-dom": "^16.12.0"
   },
   "dependencies": {
-    "@material-ui/core": "^4.6.0",
-    "@material-ui/icons": "4.5.1",
-    "@material-ui/styles": "^4.6.0",
-    "axios": "^0.19.0",
-    "clsx": "^1.0.4",
-    "formik": "^2.0.3",
+    "@material-ui/core": "^4.9.3",
```
Thebarda: It is not necessary to update version minor and bugfixes numbers in package.json when there is `^` in front of version number.
NPM will pick up the greatest on these numbers.

bdauria: @Thebarda I ran npm-check-updates which edited the package.json automagically

PR #180, opened 2020-02-20
feat(ui): Add StatusChip Component and fix Listing Component column failing with Hooks

bdauria:


src/Snackbar/Error/index.jsx
```diff
@@ -37,7 +37,7 @@ const ErrorSnackbar = ({ message, open, onClose }) => {
   return (
     <Snackbar
       anchorOrigin={{
-        vertical: 'top',
+        vertical: 'bottom',
```
kduret: is it the behaviour we want ? (eg: BA form)

src/StatusChip/index.tsx
```diff
@@ -0,0 +1,96 @@
...
+      clickable={clickable}
+      label={label?.toUpperCase()}
+      className={classes.chip}
+      {...rest}
+    />
+  );
+};
+
+export default StatusChip;
+export { StatusCode };
```
kduret: do we have a rule to order exports ? default one before named exports ?

src/Snackbar/Error/index.jsx
bdauria: I think we should have the same behaviors across all pages for this kind of feature

src/StatusChip/index.tsx
bdauria: We don't have any rule for that yet. What's your opinion, default should be last?

kduret: yes, IMHO, I think default should be the last one

src/Snackbar/Error/index.jsx
kduret: Yes indeed, but my question was more, "do we want it at the bottom" ?
But I guess yes :)

bdauria: Yes, the standard position is bottom (that's also what the UV wireframes show)

src/StatusChip/index.tsx
bdauria: Do you mind updating the wiki?

src/Listing/index.tsx
```diff
@@ -429,7 +421,10 @@ const Listing = ({
               })}
               {tableData.length < 1 && (
                 <ListingRow tabIndex={-1}>
-                  <BodyTableCell colSpan={6} align="center">
+                  <BodyTableCell
+                    colSpan={columnConfiguration.length + 1}
```
Thebarda: Maybe you should test if checkboxes are displayed through `checkable` before adding 1

bdauria: Having an extra colSpan doesn't differ visually, so I don't think a check is necessary.

PR #182, opened 2020-02-24
enh(Listing): Display skeleton when table is loading at first time

Thebarda:


src/Listing/Skeleton.tsx
```diff
@@ -0,0 +1,36 @@
...
+  BodyTableCell;
+}
+
+const ListingLoadingSkeleton = ({
+  columnConfiguration,
+  checkable,
+  BodyTableCell,
+}: LoadingSkeletonProps): JSX.Element => (
+  <>
+    {new Array(10).fill(0).map(() => (
```
kduret: 10 could be a prop cause we may know the number of items per page we expect

src/Listing/index.stories.tsx
```diff
@@ -79,3 +79,11 @@ const Story = (props): JSX.Element => {
 };
 
 export const normal = (): JSX.Element => <Story />;
+
+export const loadingWithNoData = (): JSX.Element => {
+  return <Story tableData={[]} totalRows={0} loading />;
+};
+
+export const loadingWithData = (): JSX.Element => {
+  return <Story loading />;
```
kduret: using wave animation does not break snapshot testing ?

bdauria: We have an error threshold that should cover this case

Thebarda: I have

PR #183, opened 2020-02-25
Refactor snackbar to display error and success message

Thebarda:


src/Snackbar/index.tsx
```diff
@@ -24,12 +29,32 @@ const useStyles = makeStyles((theme) => ({
   },
 }));
 
-const ErrorSnackbar = ({ message, open, onClose }) => {
+interface Props {
+  message?: string;
+  open: boolean;
+  onClose?: () => void;
+  isError?: boolean;
```
kduret: what do you think to put a severity prop which is an enum instead of a boolean ?
This will allow to manage other states like warning or info
And to make it consistent with material-ui : https://material-ui.com/components/snackbars/#customized-snackbars)

Thebarda: Should-we define the severity when we want to show a message ? Like :
```javascript
showMessage({ message: 'myMessage', severity: 'info'})
```

kduret: I think so.
IMHO, default severity should be "success"

src/Snackbar/index.tsx
```diff
@@ -1,19 +1,24 @@
...
+  SnackbarContent,
+  IconButton,
+  makeStyles,
+} from '@material-ui/core';
+import { Error, CheckCircle, Close } from '@material-ui/icons';
 
-import ErrorIcon from '@material-ui/icons/Error';
-import CloseIcon from '@material-ui/icons/Close';
-import { makeStyles } from '@material-ui/core/styles';
+import PropTypes from 'prop-types';
```
bdauria: TypeScript doesn't need PropTypes

src/Snackbar/index.tsx
bdauria: Sounds good, but severity should be an actual enum, not a string

src/Snackbar/index.tsx
```diff
@@ -45,27 +70,16 @@ const ErrorSnackbar = ({ message, open, onClose }) => {
       onClose={onClose}
     >
       <SnackbarContent
-        className={classes.error}
+        className={isError ? classes.error : classes.success}
```
bdauria: Can we maybe use clsx ?

src/Snackbar/withSnackbar.tsx
```diff
@@ -0,0 +1,66 @@
+import React, { createContext, ReactNode, ReactElement } from 'react';
+
+import Snackbar from '.';
+import useMessage from './useMessage';
+
+export interface SnackbarState {
+  showError: (message) => void;
+  showErrors: (errors) => void;
+  showSuccess: (message) => void;
+  showSuccesses: (successes) => void;
```
bdauria: I don't think we'll ever need that

src/Snackbar/index.tsx
Thebarda: I will refactor that to include severity system

PR #185, opened 2020-03-02
enh(Listing): Use React.memo to avoid re-render on Listing hover

Thebarda:


src/Listing/index.tsx
```diff
@@ -397,25 +397,27 @@ const Listing = ({
...
-                          }}
-                          color="primary"
-                        />
-                      </BodyTableCell>
-                    ) : null}
-
-                    {columnConfiguration.map((column) =>
-                      getColumnCell({ column, row }),
-                    )}
+                    <Row isHovered={hovered === row.id}>
```
bdauria: Can't we wrap React.memo around ListingRow instead?

src/Listing/index.tsx
```diff
@@ -437,4 +439,16 @@ const Listing = ({
   );
 };
 
+interface RowProps {
+  children;
+  isHovered?: boolean;
```
kduret: isHovered can be undefined ?

src/Listing/index.tsx
Thebarda: In this cas, we have to check when each prop changes

src/Listing/index.tsx
Thebarda: yes. For example, when we display loading / empty table message

PR #190, opened 2020-03-09
feat(ui): Implement Autocomplete and ConnectedAutocomplete

bdauria:


src/InputField/Select/Autocomplete/Connected/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+      })
+      .catch(() => setOptions([]))
+      .finally(() => setLoading(false));
+  };
+
+  const changeText = (event): void => {
+    loadOptions(getSearchEndpoint(event.target.value));
+  };
+
+  const doOpen = (): void => {
```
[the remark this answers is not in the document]
bdauria: There is a variable named open already, so I had to chose another name for the function.

bdauria:
@kduret There is no visual difference between the Autocomplete and the ConnectedAutocomplete, so in my opinion, the existing story covers it already. 

I might add a story for the loading cases though.

src/InputField/Select/Autocomplete/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+        key={option.id}
+        label={option.name}
+        size="small"
+        {...getTagProps({ index })}
+      />
+    ));
+
+  return (
+    <Autocomplete
+      multiple
```
Thebarda: Maybe in another PR but Autocomplete could not allow multiple value

src/api/useCancelTokenSource.ts
```diff
@@ -0,0 +1,11 @@
...
+
+import axios, { CancelTokenSource } from 'axios';
+
+const useCancelTokenSource = (): CancelTokenSource => {
+  const [cancelTokenSource] = useState(axios.CancelToken.source());
+
+  return cancelTokenSource;
+};
+
+export default useCancelTokenSource;
```
Thebarda: Is it the time to also export `useCancelTokenSource` from centreon-ui ?

kduret:
> @kduret There is no visual difference between the Autocomplete and the ConnectedAutocomplete, so in my opinion, the existing story covers it already.
> 
> I might add a story for the loading cases though.

Ok no problem. But IMHO, the storybook should be also a developer guide which show some code examples, even if it renders the same visual

src/InputField/Select/Autocomplete/Connected/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+  getSearchEndpoint: (searchField: string) => string;
+  getOptionsFromResult: (result) => Array<SelectEntry>;
+}
+
+const ConnectedAutocompleteField = <TData extends Record<string, unknown>>({
+  baseEndpoint,
+  getSearchEndpoint,
+  getOptionsFromResult,
+  ...props
+}: Props & AutocompleteFieldProps): JSX.Element => {
```
jdelpierre: I don't understand this syntax

bdauria: The ConnectedAutocompleteField is a 'generic' Component, with the type being expressed between the <>. If you look at the code below, you'll notive that the generic TData type is used to define the type of options received from the endpoint. More documentation here: https://www.typescriptlang.org/docs/handbook/generics.html

the Props & AutocompleteFieldProps is a union type between the declared Props type (declared above) and the actuel Autocomplete Props (declare as AutocompleteFieldProps). What we're doing here is basically augmenting the Prop type from the Autocomplete with 3 new props.

PR #191, opened 2020-03-11
fix(ui): Fix bugs on Listing and Snackbar

Thebarda:


src/Snackbar/index.tsx
```diff
@@ -83,7 +83,7 @@ const ErrorSnackbar = ({
       }}
       open={open}
       autoHideDuration={6000}
-      onClose={onClose}
+      onExited={onClose}
```
jdelpierre: Why you change it ?

Thebarda: Because onClose is called when component will be closed, before transition animation. Thus severity will be "emptied" and display, during close transition, a success snackbar instead an error snackbar for example.

jdelpierre: Oki :)

src/Listing/index.tsx
```diff
@@ -461,14 +462,19 @@ const Listing = ({
 interface RowProps {
   children;
   isHovered?: boolean;
+  row;
 }
 
 const MemoizedRow = React.memo<RowProps & TableRowProps>(
   ({ children, ...props }: RowProps): JSX.Element => (
     <TableRow {...props}>{children}</TableRow>
```
bdauria: We shouldn't pass the 'isHovered' prop here as it's not a valid one for TableRow

PR #192, opened 2020-03-12
enh(ui): Disable next button when a field is required

Thebarda:


src/Wizard/index.jsx
```diff
@@ -155,7 +168,11 @@ const Wizard = (props) => {
                 })}
                 {!activePage.props.noActionBar && (
                   <ActionBar
-                    disabledNext={!bag.isValid || bag.isSubmitting}
+                    disabledNext={
```
bdauria: I'd extract this condition into a variable with an explicit name cause it's starting to be unclear

src/Wizard/index.jsx
```diff
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useState, useEffect } from 'react';
```
bdauria:
```suggestion
import React, { useState, useEffect } from 'react';

```

PR #193, opened 2020-03-12
fix(ui): Fix snackbar transition color when closing it

Thebarda:


bdauria:
@Thebarda 

The snackbar still changes back to success severity color when fading out:

snackbar

PR #195, opened 2020-03-16
feat(ui): Add Actions Component to Listing

bdauria:
Actions can now be embedded within the same bar as the pagination. It will always be displayed on the left, and the pagination pushed to the right.


src/Listing/index.stories.tsx
```diff
@@ -106,3 +106,11 @@ export const loadingWithNoData = (): JSX.Element => {
 export const loadingWithData = (): JSX.Element => {
   return <Story loading />;
 };
+
+const Actions = (
```
jdelpierre: Actions is "Action button" ?

PR #196, opened 2020-03-18
enh(ui): Improve Component column in Listing

bdauria:


src/Listing/index.tsx
```diff
@@ -309,43 +311,36 @@ const Listing = ({
...
+        const { Component, clickable, hiddenCondition, width } = column;
+
+        const isCellHidden = hiddenCondition
+          ? hiddenCondition({ row, isRowSelected })
+          : false;
 
-        interface CellProps {
-          children: React.ReactNode;
-          width?: number;
+        if (isCellHidden) {
```
bdauria: This is needed for BAM, within the Indicator in impact mode, when a row is unselected. In that case, we don't want to display the cell at all (rendering null), otherwise it creates additional space in the Listing which adds unnecessary x-scroll.

src/Listing/index.tsx
```diff
@@ -309,43 +311,36 @@ const Listing = ({
           )}
         </BodyTableCell>
       ),
-      [TABLE_COLUMN_TYPES.component]: (): JSX.Element => {
-        const { Component, ComponentOnHover, clickable } = column;
+      [TABLE_COLUMN_TYPES.component]: (): JSX.Element | null => {
```
jdelpierre: What is this ? (syntax  & element)

bdauria: This is a dynamic key / value syntax
```
const id = 3;

console.log({ [id]: 'foo' }); // this outputs { 3: 'foo' } 
```

JSX.Element is a (TypeScript) type for an element output by the JSX notation (a React element basically). So the signature `(): JSX.Element | null` means: a function that returns a JSX Element OR null

jdelpierre: Oh, yeah, that I read and understood the doc.

`This is the format [TABLE_COLUMN_TYPES.component]:`

The square brackets around "variable"?

bdauria: It's there to declare a dynamic key. So Since the result of TABLE_COLUMN_TYPES.component is 4 (it's an enum), it's going to be: `..., 4: {...`

PR #197, opened 2020-03-18
enh(ui): Add Dialog submitting indicator + update packages

bdauria:


src/Dialog/index.tsx
```diff
@@ -0,0 +1,66 @@
+import React from 'react';
```
jdelpierre: import * as React from 'react';

src/Dialog/index.tsx
```diff
@@ -0,0 +1,66 @@
...
+  onCancel,
+  onConfirm,
+  labelTitle,
+  labelCancel = 'Cancel',
+  labelConfirm = 'Confirm',
+  children,
+  contentWidth,
+  confirmDisabled = false,
+  submitting = false,
+  ...rest
```
jdelpierre: What is "rest" term ?

bdauria: we destructure the props that we don't explicitly use in order to forward them (here to the `MuiDialog`)

jdelpierre: Yes, that's okay, but where did "rest" come from? Cause I couldn't find a match. Is that a reserved term?

Thebarda: Maybe  `...props` is more appropriate ?

PR #205, opened 2020-03-25
enh(Listing): Add props to disable table max-height compute

Thebarda:


bdauria:
@Thebarda How does the scrolling within the Listing behaves without the max-height attribute set?

Thebarda:
> @Thebarda How does the scrolling within the Listing behaves without the max-height attribute set?

It scrolls outside the table

bdauria:
> > @Thebarda How does the scrolling within the Listing behaves without the max-height attribute set?
> 
> It scrolls outside the table

In that case, I suggest we rename the prop to: `innerScrollDisabled` to make it clear what how the Listing will behave after setting it.

PR #208, opened 2020-03-27
enh(typing): Add string to SelectEntry id

bdauria:


src/InputField/Select/index.tsx
```diff
@@ -18,16 +19,16 @@ const useStyles = makeStyles((theme: Theme) => ({
 }));
 
 export interface SelectEntry {
-  id: number;
+  id: number | string;
```
jdelpierre: Why it's possible to accept a string on ID ?

bdauria: There are cases where the API returns a string as an ID, so number is not guaranteed here

jdelpierre: Oki ;)

PR #209, opened 2020-03-28
enh(listing): update style of listing (more dense)

kduret:


src/Listing/Header.tsx
```diff
@@ -13,12 +13,16 @@ import {
 const HeaderCell = withStyles((theme) => ({
   root: {
     backgroundColor: theme.palette.common.white,
-    lineHeight: 1.4,
```
bdauria: We might want to test this in production build, as I remember setting the lineHeight because of alignment problems in production

kduret: ok, I didn't know.
I'll test it on the built docker container

kduret: @bdauria I checked with production, and alignment seems not broken (test with chrome, firefox and Microsoft Edge)

PR #211, opened 2020-04-01
enh(listing): introduce sortField parameter and rename headRows prop

kduret:
Refs: MON-5021


src/Listing/Header.tsx
```diff
@@ -64,22 +66,27 @@ const ListingHeader = ({
...
+                {column.label}
+              </HeaderTypography>
             ) : (
               <TableSortLabel
-                active={orderBy === row.id}
+                aria-label={`Column ${column.label}`}
+                active={orderBy === getSortField(column)}
                 direction={order || 'desc'}
-                onClick={createSortHandler(row.id)}
+                onClick={createSortHandler(getSortField(column))}
```
bdauria, line 85: Can we add a test to cover this new feature? we already have some in the index.test.tsx of the Listing

kduret: ok I'll add it

PR #212, opened 2020-04-01
enh(Listing): Add prop condition about disabling checkbox following row data

Thebarda:


src/Listing/index.tsx
```diff
@@ -77,6 +77,7 @@ const cumulativeOffset = (element): number => {
 
 interface Props {
   checkable?: boolean;
+  checkCondition?;
```
bdauria, line 80:
```suggestion
  disableRowCheckCondition?;
```

src/Listing/index.tsx
```diff
@@ -115,6 +116,7 @@ const Listing = ({
   currentPage = 0,
   totalRows = 0,
   checkable = false,
+  checkCondition,
```
bdauria, line 119:
```suggestion
  disableRowCheckCondition,
```

src/Listing/index.tsx
```diff
@@ -429,6 +431,9 @@ const Listing = ({
                           inputProps={{
                             'aria-label': `Select row ${row.id}`,
                           }}
+                          disabled={
+                            checkCondition ? !!checkCondition(row) : false
```
bdauria, line 435:
```suggestion
                            disableRowCheckCondition.?(row) || false
```

bdauria:
@Thebarda We might want to update the stories to add this case too

src/Listing/index.tsx
```diff
@@ -429,6 +431,7 @@ const Listing = ({
                           inputProps={{
                             'aria-label': `Select row ${row.id}`,
                           }}
+                          disabled={!!disableRowCheckCondition(row)}
```
kduret, line 434: Is it needed to cast it ?

bdauria:
@Thebarda it seems that storyshots are not updated

Thebarda:
> @Thebarda it seems that storyshots are not updated

Updating snapshots does not change anything strangely

src/Listing/index.stories.tsx
```diff
@@ -91,6 +91,7 @@ const Story = (props): JSX.Element => {
         rowColorConditions={rowColorConditions}
         selectedRows={listing.filter((row) => row.selected)}
         checkable
+        disableRowCheckCondition={(row): boolean => row.selected}
```
bdauria, line 94: I think we should set the condition to something else than the row is selected

Thebarda: Replace by what ?

bdauria:
> > @Thebarda it seems that storyshots are not updated
> 
> Updating snapshots does not change anything strangely

Probably because of the threshold.

We can merge it like this for now, and enhence the structure of the stories later on

PR #213, opened 2020-04-02
enh(ui): Adjust header padding and add width to string column

bdauria:


src/Listing/Header.tsx
```diff
@@ -13,7 +13,9 @@ import {
 const HeaderCell = withStyles((theme) => ({
   root: {
     backgroundColor: theme.palette.common.white,
-    padding: theme.spacing(1),
```
kduret, line 16: if padding-right needs to be 0 too, it could be replaced by theme.spacing(1,0)

PR #215, opened 2020-04-03
enh(Wizard): Do not display stepper when there is only one step

Thebarda:


bdauria:
@Thebarda I think that's a good time to add a unit test to cover this case

src/Wizard/index.stories.jsx
```diff
@@ -6,6 +6,16 @@ import { Wizard, WizardPage as Page } from '..';
 
 export default { title: 'Wizard' };
 
+export const oneStep = () => (
+  <Wizard open>
+    <Page>
+      <Typography variant="h5" align="center">
```
jdelpierre, line 12: It's not possible to add a "align:center" on css theme ? 

Because I think all h5s of this guy should be similar in style?

PR #217, opened 2020-04-11
enh(Listing): Clear selection when "select all" checkbox is indeterminated

Thebarda:


src/Listing/index.test.jsx
```diff
@@ -79,7 +81,7 @@ describe('Table', () => {
         onSelectRows={onSelectRows}
         columnConfiguration={columnConfiguration}
         tableData={tableData}
-        totalRows={2}
+        totalRows={4}
```
jdelpierre, line 84: Why not index 3?

kduret: Because there are 4 rows ;)

jdelpierre: Ah yes ! I don’t see correctly this change 🙄

PR #221, opened 2020-04-12
enh(UI): Reduce space inside expansion panel summary and details

Thebarda:


bdauria:
@Thebarda there is something weird in the right bar, it seems the border continues until the bottom, while the panel actually stops before (see the storyshot)

src/RightPanel/index.jsx
```diff
@@ -148,15 +148,17 @@ const RightPanel = ({
...
-          <SecondaryPanelBar
-            aria-label="Close Secondary Panel"
-            display="flex"
-            alignItems="center"
-            alignContent="center"
-            onClick={toggleSecondaryPanel}
-          >
-            {secondaryPanelActive && <CloseSecondaryPanelIcon />}
-          </SecondaryPanelBar>
+          {secondaryPanelActive && (
```
bdauria, line 151: The panel bar should always be visible, regardless of the secondary panel is active or not

jdelpierre:
> here is something weird in the right bar, it seems the border continues until the bottom, while the panel actually stops before (see the storyshot)

I have the same remark, is that normal?

Thebarda:
> > here is something weird in the right bar, it seems the border continues until the bottom, while the panel actually stops before (see the storyshot)
> 
> I have the same remark, is that normal?

I fixed that

src/RightPanel/index.jsx
```diff
@@ -57,8 +57,8 @@ const MainPanel = styled(Box)({
 const SecondaryPanelBar = styled(Box)({
   border: '1px solid #D6D6D8',
   width: 20,
-  height: '100%',
   cursor: 'pointer',
+  marginBottom: '49px',
```
bdauria, line 61: Is this supposed to correspond to the height of the HeaderContainer? in that case, we need to extract a variable to synchronize both values

bdauria, line 61: Is this supposed to correspond to the height of the HeaderContainer? in that case, we need to extract a variable to synchronize both values

PR #226, opened 2020-04-21
feat(api): Add useRequest hook

bdauria:
## Description

This adds the useRequest hook to avoid API request logic duplication (request cancellation, loading indication and error handling).
Example of basic usage:
```
import * as React from 'react';

import getEntity from '../api';
import LoadingIndicator from '../LoadingIndicator';

const MyComponent = () => {
  const [entity, setEntity] = useState({});
  const { sendRequest, sending } = useGet({
    request: getEntity,
  });

  React.useEffect(() => sendRequest().then(setEntity), []);

  if (fetching) {
    return <LoadingIndicator />
  }
  
 return <h1>{entity.name}</h1>;
}
``` 

In case of error, the hook will try to retrieve the API error message (by accessing error.response.data.message) and display it via the Snackbar. If the path is not accessible, it will default to a given error (defaultLabelFailure).  

If a custom error message dependping on the caught error must be displayed instead, the prop `getErrorMessage` can be used, which is a function accepting the error, and returing a string. Example:

```
const getErrorMessage = (error) => {
if (error.status === 404) {
  return 'Not found';
}

return 'Something went wrong...';
```


bdauria:
@Thebarda Actually I think we can already make it method agnostic, by accepting the request function externally instead of just the endpoint. I'll give it a shot

bdauria:
> @Thebarda Actually I think we can already make it method agnostic, by accepting the request function externally instead of just the endpoint. I'll give it a shot

@Thebarda done :)

src/api/useRequest/index.ts
```diff
@@ -0,0 +1,62 @@
...
+
+const useRequest = <TResult>({
+  request,
+  getErrorMessage,
+  defaultFailureMessage = 'Oops, something went wrong',
+}: RequestParams<TResult>): RequestResult<TResult> => {
+  const { token, cancel } = useCancelTokenSource();
+  const { showMessage } = useSnackbar();
+
+  const [sending, setSending] = React.useState(true);
```
kduret, line 28: sending should be true only when sendRequest is called
```suggestion
  const [sending, setSending] = React.useState(false);
```

src/api/useRequest/index.ts
```diff
@@ -0,0 +1,62 @@
...
+    showMessage({
+      message,
+      severity: Severity.error,
+    });
+  };
+
+  const sendRequest = (): Promise<TResult> => {
+    setSending(true);
+
+    return request(token)
```
Thebarda, line 49: If I understand well. request shall be implemented like that ?
```javascript
 request : (token) => get('apiurl').then((result) => console.log(result))
```

bdauria: As per typing:
``` 
 request: (token) => Promise<TResult>;
```
So for instance:
```
const listEntity = (token) => axios.get('/entityEndpoint', { cancelToken: token });
```

Thebarda: Ok but where do we implement onSuccess callback ?

bdauria: We don't. Since we're returning the sendRequest function, which is a Promise, we can handle the result directly from it (I updated the example in the PR description).

PR #229, opened 2020-04-27
Disable all action bar buttons on error

Thebarda:
This PR disables all action bar buttons (Previous, Next/Finish) where there is an error in current step form.


kduret:
IMHO, we should be able to go to the previous step if there is an error in the current step.

PR #230, opened 2020-04-28
enh(Wizard): Validate form when changing step

Thebarda:
This PR triggers manually a Formik validation when changing step.

Here is a sample of the following bug:
Centreon - IT   Network Monitoring - Google Chrome 2020-04-28 16-39-24


src/Wizard/index.test.jsx
```diff
@@ -56,4 +75,26 @@ describe('Wizard', () => {
 
     expect(getByText('Step 1')).toBeInTheDocument();
   });
+
+  it('can not click on "Finish" button when second step has an error, goes to first step by clicking on "Previous" then return to second step', async () => {
```
bdauria, line 79:
```suggestion
  it('cannot finish the Wazard when there is a validation error, but can change steps', async () => {
```

Test names should describe the case, not describe the details of the steps in place

src/Wizard/index.jsx
```diff
@@ -28,6 +28,14 @@ const cloneElement = (element, props) => {
   return React.cloneElement(element, { ...forwardedProps });
 };
 
+const ValidateForm = ({ children, page, validateForm }) => {
```
bdauria, line 31:
```suggestion
const FormPage = ({ children, page, validateForm }) => {
```

Thebarda: That was the first name of this component x)

package.json
```diff
@@ -47,7 +47,8 @@
     "react-files": "^2.4.8",
     "react-router-dom": "^5.1.2",
     "resize-observer-polyfill": "^1.5.1",
-    "use-debounce": "^3.4.1"
+    "use-debounce": "^3.4.1",
+    "yup": "^0.28.4"
```
kduret, line 51: should we move yup as a dev dependency since it's only used in test files

PR #231, opened 2020-04-29
fix(UI): Fix custom labels display on action bar

Thebarda:


src/Wizard/ActionBar.test.jsx
```diff
@@ -20,4 +22,40 @@ describe('ActionBar', () => {
 
     expect(getByText('Finish').parentNode).toHaveAttribute('disabled');
   });
+
+  it('displays custom previous, next button label when wizar is on second step and not on last step', () => {
```
bdauria, line 26:
```suggestion
  it('displays the given previous and next labels when the current page is not the last one', () => {
```

src/Wizard/ActionBar.test.jsx
```diff
@@ -20,4 +22,40 @@ describe('ActionBar', () => {
...
+        labelPrevious="Custom previous"
+        labelNext="Custom next"
+      />,
+    );
+
+    expect(getByText('Custom previous')).toBeInTheDocument();
+    expect(getByText('Custom next')).toBeInTheDocument();
+  });
+
+  it('displays custom previous, finish button label when wizard is on last step', () => {
```
bdauria, line 40:
```suggestion
  it('displays the given previous and finish labels when the current page is the last one', () => {
```

PR #233, opened 2020-05-13
PoC(ts.data.json): Add decoder to request custom hook

Thebarda:
This PR integrates JSON API decoder to request custom hook.
It also logs errors from API request and decoding using [ulog](https://github.com/Download/ulog)

This PR is related with https://github.com/centreon/centreon-autodiscovery/pull/258

### DIFF
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -3,4 +3,31 @@ import axios from 'axios';
 const getData = <TData>(cancelToken) => (endpoint): Promise<TData> =>
   axios.get(endpoint, { cancelToken }).then(({ data }) => data);
 
-export { getData };
+const postData = <TData>(cancelToken) => ({
+  endpoint,
+  datas,
+}): Promise<TData> =>
+  axios
+    .post(endpoint, datas, {
+      headers: { 'Content-Type': 'application/x-www-form-urlencoded' },
+      cancelToken,
+    })
+    .then(({ data }) => data);
+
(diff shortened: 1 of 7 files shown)


src/api/useRequest/index.ts
```diff
@@ -48,12 +54,20 @@ const useRequest = <TResult>({
...
-      .catch((error) =>
-        cond([
+      .then((data) => {
+        if (decoder) {
+          return decoder.decodePromise(data);
+        }
+        return data;
+      })
+      .catch((error) => {
+        log.error(error);
```
bdauria, line 64: How does the log look like with a real-world error?

Thebarda: paginationError

Here is a sample when property 'page' from 'pagination is not a number

src/api/index.ts
```diff
@@ -3,4 +3,31 @@ import axios from 'axios';
...
   axios.get(endpoint, { cancelToken }).then(({ data }) => data);
 
-export { getData };
+const postData = <TData>(cancelToken) => ({
+  endpoint,
+  datas,
+}): Promise<TData> =>
+  axios
+    .post(endpoint, datas, {
+      headers: { 'Content-Type': 'application/x-www-form-urlencoded' },
```
jdelpierre, line 12: it's not possible to create a "constant" of headers ? Because looks like it's the same for every type of call.

PR #234, opened 2020-05-19
feat(UI): Add search help tooltip

Thebarda:
This PR adds same search input help tooltip as events view because it will be use on modules as HD.
searchHelpTooltip


src/InputField/Search/SearchHelpTooltip.tsx
```diff
@@ -0,0 +1,127 @@
...
+      </Box>
+    </>
+  );
+};
+
+interface TooltipProps extends ContentProps {
+  labelSearchHelp: string;
+}
+
+const SearchHelpTooltip = ({
```
bdauria, line 74: I like the idea to make this re-usable, but I would rename it to something more generic as it can also be useful for some other kind of tooltips. What about PersistentTooltip?

src/InputField/Search/PersistentTooltip.tsx
```diff
@@ -0,0 +1,142 @@
...
+        {displayElement(
+          <i>
+            <b>{`${labelTips}: `}</b>
+            {`${labelGetHelp} `}
+            <Link href={urlTip} target="_blank" rel="noopener noreferrer">
+              {labelUrlTip}
+            </Link>
+          </i>,
+        )(labelTips && labelGetHelp && urlTip && labelUrlTip)}
+      </Box>
```
bdauria, line 73: I'm wondering if we want to be that enclosed in the implementation. Wouldn't it be better to let the implementation of the content to the calling Component?

src/InputField/Search/PersistentTooltip.tsx
```diff
@@ -0,0 +1,74 @@
...
+  const title = (
+    <>
+      <IconButton
+        size="small"
+        onClick={closeTooltip}
+        className={classes.buttonClose}
+      >
+        <IconClose fontSize="small" />
+      </IconButton>
+      {content}
```
bdauria, line 52: a children prop would be more appropriate here

Thebarda:
> IMHO, we should add the possibility to set this tooltip as controlled, like it's done in material-ui's tooltip with useControlled hook (https://github.com/mui-org/material-ui/blob/master/packages/material-ui/src/Tooltip/Tooltip.js#L203)
> This will allow to test it with storyshots. Cause current generated storyshots do not display tooltip

So, you want to add a `useTooltip` that control `open`state and pass that `open` to PersistentTooltip as prop, isn't it ?

kduret:
> > IMHO, we should add the possibility to set this tooltip as controlled, like it's done in material-ui's tooltip with useControlled hook (https://github.com/mui-org/material-ui/blob/master/packages/material-ui/src/Tooltip/Tooltip.js#L203)
> > This will allow to test it with storyshots. Cause current generated storyshots do not display tooltip
> 
> @kduret So, you want to add a `useTooltip` that control `open`state and pass that `open` to PersistentTooltip as prop, isn't it ? And would you use both uncontrolled and controlled tooltip ? Or maybe just add a `defaultOpen` prop ?
> 
> Note : It is difficult to test an opened tooltip because it triggers an animation and jest snapshot is triggered before that animation

my suggestion was just to set a `open` prop to persistentTooltip component. This prop should be optional. Its purpose is to put the component as controlled. But maybe it's a bad idea et make it too complex. You should decide @bdauria and @Thebarda

Thebarda:
> > > IMHO, we should add the possibility to set this tooltip as controlled, like it's done in material-ui's tooltip with useControlled hook (https://github.com/mui-org/material-ui/blob/master/packages/material-ui/src/Tooltip/Tooltip.js#L203)
> > > This will allow to test it with storyshots. Cause current generated storyshots do not display tooltip
> > 
> > 
> > @kduret So, you want to add a `useTooltip` that control `open`state and pass that `open` to PersistentTooltip as prop, isn't it ? And would you use both uncontrolled and controlled tooltip ? Or maybe just add a `defaultOpen` prop ?
> > Note : It is difficult to test an opened tooltip because it triggers an animation and jest snapshot is triggered before that animation
> 
> my suggestion was just to set a `open` prop to persistentTooltip component. This prop should be optional. Its purpose is to put the component as controlled. But maybe it's a bad idea et make it too complex. You should decide @bdauria and @Thebarda

My bad, it snapshots correctly opened tooltips

bdauria:
@Thebarda @kduret I find it a bit weird to have an open prop just for the sake of testing, but if there is no other way around it, then we should go for it.

src/InputField/Search/PersistentTooltip.tsx
```diff
@@ -0,0 +1,92 @@
...
+}: Props): JSX.Element => {
+  const classes = useStyles();
+
+  const [open, setOpen] = React.useState(openTooltip || false);
+
+  const toggleTooltip = (): void => {
+    ifElse(
+      isNil,
+      () => setOpen(!open),
+      () => toggleTooltipProp && toggleTooltipProp(),
```
bdauria, line 47: I think we would gain in readability if we use cond() instead of ifElse() since we actually have 3 cases, rather than 2

src/InputField/Search/PersistentTooltip.tsx
```diff
@@ -0,0 +1,92 @@
...
+
+  const closeTooltip = (): void => {
+    if (openTooltip) {
+      if (closeTooltipProp) {
+        closeTooltipProp();
+      }
+    } else {
+      setOpen(false);
+    }
+  };
```
bdauria, line 59:
```suggestion
    if (!openTooltip) {
     setOpen(false);
     return;
   }
   
   closeTooltipProp.?();
  };
```

PR #235, opened 2020-05-22
enh(Autocomplete): Add end adornment to Autocomplete

Thebarda:
singleAdornment
singleAdornmentMulti

PR #236, opened 2020-05-27
chore(npm): Update packages and update resulting typing errors

bdauria:


src/InputField/Select/Autocomplete/Multi/index.tsx
```diff
@@ -20,7 +20,10 @@ export type Props = Omit<
   AutocompleteProps,
   'renderTags' | 'renderOption' | 'multiple'
 > &
-  Omit<UseAutocompleteMultipleProps<SelectEntry>, 'multiple'>;
+  Omit<
+    MuiAutocompleteProps<SelectEntry, boolean, boolean, boolean>,
```
Thebarda, line 24: This line will confilct with #235

src/Snackbar/withSnackbar.tsx
```diff
@@ -21,7 +21,7 @@ interface SnackbarContextProviderProps {
   children?: ReactNode;
 }
 
-const withSnackbar = (Component): ((props) => ReactElement) => {
+const withSnackbar = (Component: React.SFC): ((props) => ReactElement) => {
```
jdelpierre, line 24: What is the SFC ?
I see this : https://stackoverflow.com/questions/53885993/react-16-7-react-sfc-is-now-deprecated
it's same ?

PR #237, opened 2020-05-27
fix(autocomplete): Fix single autcomplete value selection

Thebarda:
This PR fixes single connected autocomplete value selection.

singleconnectedvalue


src/InputField/Select/Autocomplete/Connected/index.tsx
```diff
@@ -64,14 +68,20 @@ export default (
 
     const loading = sending || !options;
 
+    const inputValueProps: InputValueProps = {};
+
+    if (optionsOpen) {
+      inputValueProps.inputValue = searchValue;
+    }
+
```
bdauria, line 76:
```suggestion
    const inputValueProps = optionsOpen ? { inputValue: searchValue } : {};

```

PR #238, opened 2020-06-02
enh(typing): Simplify error prop for TextField and add error prop to SelectField

bdauria:
This add an error prop to the SelectField and makes things consistent and easier by changing the error prop of TextField to string instead of boolean


kduret:
I'm not sure this storyshot should be displayed as an error : 
image

PR #239, opened 2020-06-05
enh(UI): Prevent enter key pressed in Wizard

Thebarda:
This PR prevents enter key pressed in whole wizard.
Indeed, when we click on next to go to second step, click again on next to go to third step then click on previous to go to second step and finally press Enter key, it goes to first step. Because Previous step button seems to keep like a focus.


Thebarda:
> is this feature of deleting the "enter" key really intended ?
> Will this change have to be applied to all wizards to make it uniform everywhere ?

This will change all wizard behavior about preventing "enter" key. Currently, this wizard is used only in Autodiscovery and License Manager.
In fact, by pressing "Enter" when there is a focus in a input field will go to next step. But when there is no focus inside the step it will not go to next or previous step.

PR #240, opened 2020-06-10
feat(UI): Create skeleton listing page

Thebarda:
listingPage1
listingPage2
listingPage3

Storybook_Trim
ListingPahe


src/Listing/index.tsx
```diff
@@ -353,9 +353,9 @@ const Listing = ({
   const emptyRows = limit - Math.min(limit, totalRows - currentPage * limit);
 
   const tableMaxHeight = (): string => {
-    return innerScrollDisabled
-      ? '100%'
-      : `calc(100vh - ${tableTopOffset}px - 25px)`;
+    return !innerScrollDisabled
+      ? `calc(100vh - ${tableTopOffset}px - 100px)`
+      : '100%';
```
bdauria, line 358:
```suggestion
    if (innerScrollDisabled) {
      return '100%';
    }
    
    return `calc(100vh - ${tableTopOffset}px - 100px)`;
```

I don't remember why we have to retract either 25px or 100px, but we should document it via a variable.

src/Listing/index.tsx
```diff
@@ -76,6 +76,8 @@ const cumulativeOffset = (element): number => {
   return cumulativeOffset(element.offsetParent) + element.offsetTop;
 };
 
+const heightOffset = '100px';
```
bdauria, line 79: Can we somehow get this value dynamically?

src/ListingPage/Filters.tsx
```diff
@@ -0,0 +1,93 @@
...
+  expandableFilters?: React.ReactElement;
+  expandTransitionFinished?: () => void;
+}
+
+const Filters = ({
+  filtersExpandable,
+  labelFiltersIcon,
+  filters,
+  expandableFilters,
+  expandTransitionFinished,
```
bdauria, line 60:
```suggestion
  onExpandTransitionFinish,
```
expandTransitionFinished sounds like a boolean value

src/ListingPage/Filters.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const classes = useStyles();
+
+  const toggleExpanded = () => setExpanded(!expanded);
+
+  return (
+    <div className={classes.filters}>
+      <ExpansionPanel
+        square
+        expanded={!filtersExpandable ? false : expanded}
+        onTransitionEnd={() => expandTransitionFinished?.()}
```
bdauria, line 72:
```suggestion
        onTransitionEnd={() => onExpandTransitionFinish?.()}
```

src/ListingPage/Filters.tsx
```diff
@@ -0,0 +1,93 @@
...
+  const [expanded, setExpanded] = React.useState(false);
+  const classes = useStyles();
+
+  const toggleExpanded = () => setExpanded(!expanded);
+
+  return (
+    <div className={classes.filters}>
+      <ExpansionPanel
+        square
+        expanded={!filtersExpandable ? false : expanded}
```
bdauria, line 71:
```suggestion
        expanded={filtersExpandable ? expanded : false}
```

src/ListingPage/SlidePanel.tsx
```diff
@@ -0,0 +1,46 @@
...
+    <Slide
+      direction="left"
+      in={equals(open, true)}
+      timeout={{
+        enter: 150,
+        exit: 50,
+      }}
+    >
+      <div className={classes.rightPanel}>
+        <Paper elevation={5} className={classes.paperPanel}>
```
bdauria, line 38: We need the fixed header here as it's a common use case

src/ListingPage/SlidePanel.tsx
```diff
@@ -0,0 +1,62 @@
...
+          {slideHeader && (
+            <Paper elevation={3} className={classes.slideHeader}>
+              {slideHeader}
+            </Paper>
+          )}
+          <div className={classes.slideContent}>{slidePanel}</div>
```
bdauria, line 55: In the case of the event view, the panel content is stateful, so we might want to extract this part into a dedicated Component extensible using both header and body sub-Components.

Then, instead of passing slidePanelHeader and slidePanel from above, we would just pass something slidePanelContent encapsulating this part.

src/ListingPage/index.stories.tsx
```diff
@@ -218,9 +219,16 @@ const DetailsPanelHeader = (): JSX.Element => (
...
+  <SlidePanel
+    header={<DetailsPanelHeader />}
+    content={<DetailsPanelContent />}
+  />
+);
+
 export const normal = (): JSX.Element => (
   <ListingPage
-    slidePanelOpen={false}
+    openSlidePanel={false}
```
bdauria, line 231: I think the previous name made more sense for a variable (slidePanelOpen => noun versus openSlidePanel => verb)

src/ListingPage/index.stories.tsx
```diff
@@ -0,0 +1,330 @@
...
+    slidePanelOpen
+    listing={listing}
+    filtersExpandable
+    filters={<FiltersSummary />}
+    expandableFilters={<FiltersDetails />}
+    slidePanel={<DetailsPanel />}
+  />
+);
+
+export const NormalWithFiltersDetailsAndOpenedIntegratedPanel = (): JSX.Element => {
```
bdauria, line 315:
```suggestion
export const normalWithFiltersDetailsAndOpenedIntegratedPanel = (): JSX.Element => {
```

Thebarda: Eslint requires a capital as first character because this component is using React hooks

bdauria: I'd add a comment to disable the rule in that case to stay consistent with the other stories declaration

PR #241, opened 2020-06-10
enh(ui): Cleanup Listing and add compact prop to Select and Text Fields

bdauria:
This PR cleans up a bit the Listing (as well as the BoundingBox) Components to remove some unused code, and add some improvements on column rendering.

It also adjusts the Select and Text Fields to add a compact prop in order to render them in a tiny way. Finally, the Select Field also handles menu colors whenever a color field is provided within the select entries:

Screenshot from 2020-06-10 19-03-40
Screenshot from 2020-06-10 19-03-12
Screenshot from 2020-06-10 19-03-28
localhost_4000_centreon_configuration_bam_bas_id=1


src/BoundingBox/index.js
```diff
@@ -8,8 +8,7 @@
 /* eslint-disable react/require-default-props */
 /* eslint-disable no-unused-vars */
 
-import React, { Component, Children } from 'react';
-import { findDOMNode } from 'react-dom';
+import { Component, Children } from 'react';
```
kduret, line 11: import * as React from 'react';

src/InputField/Select/index.tsx
```diff
@@ -39,24 +46,37 @@ const SelectField = ({
   selectedOptionId,
   label,
   error,
+  fullWidth,
+  ariaLabel,
+  compact = false,
```
kduret, line 51: should not be better to use the `size` prop with `xsmall` value ?

bdauria: xsmall doesn't exist as a value for size in FormControl.

compact is a very specific prop that I hope we'll get rid of as soon as we find a better way to handle Indicators selection.

PR #242, opened 2020-06-16
feat(ui): Add IconButton, remove AddButton, add withThemeProvider HOC

bdauria:
This adds the IconButton Component used to render a Button using an Icon and including a tooltip with the title.

The AddButton is also removed since it's not used anymore. 

the withThemeProvider HOC is added to ease the use of composition for reusable Components.


src/InputField/Select/Autocomplete/index.tsx
```diff
@@ -97,12 +99,14 @@ const AutocompleteField = ({
             ...params.InputProps,
             endAdornment: (
               <>
-                <InputAdornment
-                  classes={{ root: classes.inputEndAdornment }}
-                  position="end"
-                >
-                  {endAdornment}
-                </InputAdornment>
+                {endAdornment && (
```
bdauria, line 102: This prevents mui from displaying a warning regarding a change in controlling state of the Autocomplete

PR #244, opened 2020-07-02
Fix(Listing): Reset page to 0 when when changing page limit

Thebarda:
ListingPageNormal

### DIFF
--- a/src/Listing/index.test.jsx
+++ b/src/Listing/index.test.jsx
@@ -36,6 +36,35 @@ describe('Table', () => {
   const onSelectRows = jest.fn();
   const onSort = jest.fn();
 
+  const oneHundredElements = new Array(100).fill(0);
+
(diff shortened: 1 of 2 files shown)


cgagnaire:
Refs MON-5679 😗

Thebarda:
> A unit test could be added to validate it

@kduret Done

src/Listing/index.test.jsx
```diff
@@ -167,4 +196,18 @@ describe('Table', () => {
       orderBy: columnWithSortField.sortField,
     });
   });
+
+  it('resets page number to 0 after selecting a new pagination limit when current page is 4', () => {
```
bdauria, line 200:
```suggestion
  it('resets the page number when changing the limit and the current page is different than 0', () => {
```

PR #245, opened 2020-07-03
feat(Autocomplete): Add infinite scroll autocomplete

Thebarda:
single
multi


src/InputField/Select/Autocomplete/InfiniteScroll/index.tsx
```diff
@@ -0,0 +1,146 @@
...
+      }
+
+      if (optionsOpen) {
+        loadOptions({
+          endpoint: getEndpoint({ search: searchValue, page }),
+          loadMore: page > 1,
+        });
+      } else if (!optionsOpen) {
+        setOptions([]);
+        setPage(initialPage);
```
bdauria, line 125: I'm not sure this block will ever be reachable

src/utils/useObserver.ts
```diff
@@ -0,0 +1,41 @@
...
+      }
+      if (loading) {
+        observer.current = null;
+      } else {
+        observer.current = new IntersectionObserver((entries) => {
+          if (entries[0].isIntersecting && page < maxPage) {
+            action();
+          }
+        });
+      }
```
bdauria, line 30:
```suggestion
        observer.current = null;
        return;
      }
      observer.current = new IntersectionObserver((entries) => {
        if (entries[0].isIntersecting && page < maxPage) {
          action();
        }
      });
      
```

src/InputField/Select/Autocomplete/InfiniteScroll/index.tsx
```diff
@@ -0,0 +1,143 @@
...
+        return;
+      }
+
+      if (optionsOpen) {
+        loadOptions({
+          endpoint: getEndpoint({ search: searchValue, page }),
+          loadMore: page > 1,
+        });
+      }
```
bdauria, line 123:
```suggestion
      loadOptions({
        endpoint: getEndpoint({ search: searchValue, page }),
        loadMore: page > 1,
      });
```

Thebarda:
> Is it possible to add a test on infinite scroll ? (loading of page 2)

We can not test that because Jest tests on generated code not but not graphically like Cypress.

kduret:
Ok, event if I'm not totally sure this should be validated with Cypress :)
for example, material-ui validate scroll on tabs component : https://github.com/mui-org/material-ui/blob/next/packages/material-ui/src/Tabs/Tabs.test.js#L371

Thebarda:
> Ok, even if I'm not totally sure this should be validated with Cypress :)
> for example, material-ui validate scroll on tabs component : https://github.com/mui-org/material-ui/blob/next/packages/material-ui/src/Tabs/Tabs.test.js#L371

Ok I will do that later

Thebarda:
> > Ok, even if I'm not totally sure this should be validated with Cypress :)
> > for example, material-ui validate scroll on tabs component : https://github.com/mui-org/material-ui/blob/next/packages/material-ui/src/Tabs/Tabs.test.js#L371
> 
> Ok I will do that later even if I have no idea how to do that...

@kduret I looked out a little to Tabs implementation about scroll event. 
It seems they can manage scrollable tabs : https://material-ui.com/components/tabs/#scrollable-tabs
With a `onScroll` listener : https://github.com/mui-org/material-ui/blob/next/packages/material-ui/src/Tabs/Tabs.js#L488
With that they can fully control scroll state.
In our case, scroll bar is uncontrolled. We only check if last element appears on screen when scrolling. 
Both uses cases are completly different. 
Furthermore, we mock IntersectionObserver for our tests. So.... IntersectionObserver is not triggered.

During the implementation, I tried to test that feature but without success because of these reasons

kduret:
Ok, thanks for investigation ! this is much clearer now :)

jdelpierre:
I think we should post a loader for the slightly long response times. 
We should show that we have finished loading everything (end of the number of results).

PR #249, opened 2020-07-16
feat(ui): Add ContentWithCircularLoading

bdauria:
This PR adds a Component to handle some loading indication within a page.


src/ContentWithCircularProgress/index.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { CircularProgress, makeStyles } from '@material-ui/core';
+
+const useStyles = makeStyles(() => ({
+  alignCenter: {
+    width: '100%',
+    height: '100%',
+    position: 'absolute',
+    top: '50%',
+    left: '50%',
+  },
```
Thebarda, line 14: I think it's clearer to use "grid" method instead of "position: absolute" method to center an element.
Here is a [Codesandbox](https://codesandbox.io/s/center-circular-65c7x?file=/demo.js) to see the result

```suggestion
  alignCenter: {
    width: '100%',
    height: '100%',
    display: 'grid',
    justifyContent: 'center',
    alignItems: 'center',
  },
```

PR #255, opened 2020-07-30
fix(ListingPage): Adjust ListingPage to address some layout issues

bdauria:
This PR addresses a few layout issues encountered after migrating the Events view page to the ListingPage.


src/ListingPage/SlidePanel.tsx
```diff
@@ -1,32 +1,70 @@
...
+
+const useStyles = makeStyles((theme) => ({
+  container: {
+    height: '100%',
+    display: 'grid',
+    gridTemplate: 'auto auto 1fr / 1fr',
+  },
+  header: {
+    gridArea: '1 / 1 / 2 / 1',
+    padding: 8,
```
kduret, line 22:
```suggestion
    padding: theme.spacing(2),
```

src/ListingPage/SlidePanel.tsx
```diff
@@ -38,16 +76,32 @@ const SlidePanel = ({ header, content }: SlidePanelProps): JSX.Element => {
...
+          <AppBar position="static" color="default">
+            {!isEmpty(tabs) && (
+              <Tabs
+                variant="fullWidth"
+                value={selectedTabId}
+                indicatorColor="primary"
+                textColor="primary"
+                onChange={onTabSelect}
+              >
+                {tabs.map((tab) => tab)}
```
kduret, line 96: No need to add key ?

bdauria: There is going to be a linting error if the Component using the SlidePanel dosen't provide keys to the given tabs.

src/ListingPage/index.stories.tsx
```diff
@@ -241,84 +257,90 @@ interface PanelProps {
...
-      filtersExpandable
-      filters={
-        <FiltersSummaryWithOpenSlidePanelButton onOpen={() => setOpen(true)} />
-      }
-      expandableFilters={<FiltersDetails />}
-      slidePanel={<DetailsPanel onClose={() => setOpen(false)} />}
-      slidePanelIntegrate
+      filters={<ExpandableFiltersWithOpenButton onOpen={() => setOpen(true)} />}
+      panel={<DetailsPanel onClose={() => setOpen(false)} />}
+      panelFixed={false}
```
Thebarda, line 343:
```suggestion
      panelFixed
```
A story with fixed panel is missing

src/index.ts
```diff
@@ -49,7 +49,8 @@ export {
 } from './Listing/models';
 
 export { default as ListingPage } from './ListingPage';
-export { default as SlidePanel } from './ListingPage/SlidePanel';
+export { default as FilteBar } from './ListingPage/Filters';
```
Thebarda, line 52:
```suggestion
export { default as Filters } from './ListingPage/Filters';
```

src/ListingPage/index.stories.tsx
bdauria: > I see a regression about Listing Page resizing when expanding or collapsing filters
> ListingPage is no longer resizing

Can you elaborate on the regression? it seems to work as before for me

Thebarda: Before Listing Page height was resized 'ouf of the box' (see : https://github.com/centreon/centreon-ui/blob/master/src/ListingPage/index.tsx#L60)
I will check on AutoDiscovery if it actually works

src/Panel/index.tsx
```diff
@@ -0,0 +1,129 @@
...
+  id: number;
+}
+
+interface Props {
+  header: React.ReactElement;
+  selectedTab: React.ReactElement;
+  tabs?: Array<JSX.Element>;
+  selectedTabId?: number;
+  onTabSelect?: (event, id: number) => void;
+  onClose: () => void;
```
Thebarda, line 64: `onClose` is not required. For example, in hosts listing, we always display mappers panel

PR #260, opened 2020-08-06
enh(Connected/Autocomplete): Add paginationPath prop

Thebarda:
As some endpoints does not send pagination properties in the same way, this PR adds a prop to customize path to meta pagination properties.


bdauria:
@Thebarda did you test with a data structure without the pagination object?

Thebarda:
> @Thebarda did you test with a data structure without the pagination object?

@bdauria Done

src/InputField/Select/Autocomplete/Connected/index.stories.tsx
```diff
@@ -44,26 +55,56 @@ mockedAxios
...
+    getEndpoint={(params) => getEndpoint({ endpoint: baseEndpoint, params })}
     getOptionsFromResult={(result): Array<SelectEntry> => result}
     placeholder="Type here..."
   />
 );
 
 export const multi = (): JSX.Element => (
   <MultiConnectedAutocompleteField
-    label="Multi Infinite Autocomplete"
+    label="Multi Connedted Autocomplete"
```
kduret, line 90:
```suggestion
    label="Multi Connected Autocomplete"
```

PR #272, opened 2020-09-23
[Soft breaking] enh(Listing/performance): Memoize Column Cell to spare renders on Listing update

bdauria:
This is an attempt to fixe some lags appearing on columns containing hoverable Components with large listings. To do so, there is an extensive check on Props modification within the ColumnCell Component (which is now a standalone, memo Component). 

The Column models contains a few more properties necessary to make sure the Listing is still rendered correctly:

- hasHoverableComponent (boolean): indicates whether the column has a Component that changes when hovered. This is necessary to trigger a re-render when the mouse hovers the corresponding row
- renderComponentOnRowUpdate (boolean): indicates whether the (Component) column should re-render when the corresponding row (Entity) has changes.

### DIFF
--- a/src/Listing/ColumnCell.tsx
+++ b/src/Listing/ColumnCell.tsx
@@ -127,9 +127,52 @@ const ColumnCell = React.memo<Props>(
     return cellByColumnType[column.type]();
   },
   (prevProps, nextProps) => {
+    const previousColumn = prevProps.column;
+    const previousRow = prevProps.row;
+    const previousIsRowHovered = prevProps.isRowHovered;
+    const previousIsRowSelected = prevProps.isRowSelected;
+
+    const nextColumn = nextProps.column;
+    const nextRow = nextProps.row;
+    const nextIsRowHovered = nextProps.isRowHovered;
+    const nextIsRowSelected = nextProps.isRowSelected;
+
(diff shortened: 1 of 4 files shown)


Thebarda (dismissed):

src/Listing/ColumnCell.tsx
```diff
@@ -0,0 +1,189 @@
+/* eslint-disable react/prop-types */
```
kduret, line 1: do you need to disable this rule on a `tsx` file ?

src/Listing/ColumnCell.tsx
```diff
@@ -0,0 +1,189 @@
...
+
+import {
+  makeStyles,
+  withStyles,
+  TableCell,
+  Tooltip,
+  Typography,
+} from '@material-ui/core';
+
+import clsx from 'clsx';
```
kduret, line 15: move it with ramda import

src/Listing/ColumnCell.tsx
bdauria: it seems that we disabled the rule in @frontend/core, I'll check centreon-ui needs an update.

kduret (approved):

Thebarda (approved):

PR #282, opened 2020-10-27
feat(utils): Add useLocaleDateTimeFormat hook

bdauria:
This adds a hook to format dates and times using the user locale and timezone.

### DIFF
--- a/src/utils/useLocaleDateTimeFormat/index.test.ts
+++ b/src/utils/useLocaleDateTimeFormat/index.test.ts
@@ -2,11 +2,7 @@ import dayjs from 'dayjs';
 
 import 'dayjs/locale/en';
 
-import {
-  renderHook,
-  RenderHookResult,
-  act,
-} from '@testing-library/react-hooks';
+import { renderHook, RenderHookResult } from '@testing-library/react-hooks';
 import timezonePlugin from 'dayjs/plugin/timezone';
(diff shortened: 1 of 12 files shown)


Thebarda (approved):

src/utils/useLocaleDateTimeFormat/index.test.ts
```diff
@@ -0,0 +1,75 @@
...
+}));
+
+const renderUseLocaleDateTimeFormat = (): RenderHookResult<
+  unknown,
+  LocaleDateTimeFormat
+> => {
+  return renderHook(() => useLocaleDateTimeFormat());
+};
+
+const dateTime = '1995-12-17T03:24:00';
```
jdelpierre, line 33: What is the behavior for a time in afternoon ? (15:00) => 3:00PM ?

bdauria: can you point me to the doc you're mentioning?

bdauria: @jdelpierre we use HH:mm as a default format for now, we might extend it later to use another one, but we can always use the format function and pass a custom format.

Thebarda (approved):

jdelpierre (approved):

PR #284, opened 2020-10-30
enh(utils): Add humanized duration function

Thebarda:
It adds a function that returns a humanized duration (6 hours, 9 minutes and 1 second) following the current locale.


jdelpierre:
Isn't it interesting to use the "Fallback" system if the language doesn't exist? 
https://github.com/EvanHahn/HumanizeDuration.js => fallback part

Thebarda:
> Isn't it interesting to use the "Fallback" system if the language doesn't exist?
> https://github.com/EvanHahn/HumanizeDuration.js => fallback part
> 
> And add a test about it ?

I don't know how to mock @centreon/ui-context for a uniq tests suite

src/utils/useLocaleDateTimeFormat/index.ts
```diff
@@ -58,7 +68,29 @@ const useLocaleDateTimeFormat = (): LocaleDateTimeFormat => {
...
   };
 
-  return { format, toDateTime, toDate, toTime, toIsoString };
+  const toHumanizedDuration = ({
+    duration,
+    labelConjunction,
+  }: HumanizeDuration): string => {
+    const normalizedLocale = locale.substring(0, 2).toLowerCase();
+
+    return humanizeDuration(duration * 1000, {
```
bdauria, line 77: Since we're going to use this primarily in listings, I suggest we use the short version, like 'shortEn', 'shortFr' ....

Thebarda: shortEn, shortFr, shortEs, shortPt ?

Thebarda: @bdauria To gain even mor, should we have a conjunction word like "and" ?
The result is "6 h, 9 m and 1 s" for now but without the conjunction, it gonna be "6 h, 9 m, 1 s"

bdauria: that sounds good. Can we get rid of the ',' as well?

Thebarda: ofc

kduret (approved):

bdauria (approved):

PR #289, opened 2020-11-30
enh(Autocomplete/Draggable): Manage error and required on draggable

Thebarda:
It manages error and required properties on Draggable autocomplete


bdauria (dismissed):

src/InputField/Select/Autocomplete/Draggable/index.tsx
```diff
@@ -49,6 +49,8 @@ const useStyles = makeStyles((theme) => ({
 interface Props {
   onSelectedValuesChange?: (values: Array<SelectEntry>) => Array<SelectEntry>;
   label?: string;
+  labelError?: string;
```
bdauria, line 52:
```suggestion
  error?: string;
```

src/InputField/Select/Autocomplete/Draggable/index.tsx
```diff
@@ -57,9 +59,14 @@ const DraggableAutocomplete = (
   const InnerDraggableAutocompleteField = ({
     onSelectedValuesChange,
     label,
+    labelError,
```
bdauria, line 62:
```suggestion
    error,
```

src/InputField/Select/Autocomplete/Draggable/index.tsx
```diff
@@ -179,7 +186,15 @@ const DraggableAutocomplete = (
         freeSolo
         handleHomeEndKeys
         renderTags={renderTags}
-        renderInput={(params) => <TextField {...params} label={label} />}
+        renderInput={(params) => (
+          <TextField
+            {...params}
+            label={label}
+            helperText={labelError}
```
bdauria, line 193:
```suggestion
            helperText={error}
```

src/InputField/Select/Autocomplete/Draggable/index.tsx
```diff
@@ -179,7 +186,15 @@ const DraggableAutocomplete = (
...
+            error={labelError}
```
bdauria, line 194:
```suggestion
            error={error}
```

bdauria (commented):
Shouldn't the same behaviour apply to other Autocomplete as well?

kduret (dismissed):

jdelpierre (approved):

src/InputField/Select/Autocomplete/Draggable/index.tsx
jdelpierre: @bdauria why do you want to remove the word "label"? It still allows you to see quickly that we are talking about the error text, right?

bdauria (commented):
See review comments

src/InputField/Select/Autocomplete/Draggable/index.tsx
bdauria: because we use error everywhere else, including on our TextFields. The wording is correct, it's just a matter of consistency.

jdelpierre: Ok indeed in the other components we use "error" as a term ;)
But I think "labelError" was a good idea to identify more easily.

jdelpierre (dismissed):

jdelpierre (dismissed):

bdauria (commented):
@Thebarda So basically the other kinds of Autocomplete don't manage any error? can we find a way to reuse the implemented behavior across them?

Thebarda:
> 
> 
> @Thebarda So basically the other kinds of Autocomplete don't manage any error? can we find a way to reuse the implemented behavior across them?

@bdauria Every other Autocompletes manage error case

kduret (dismissed):

src/InputField/Select/Autocomplete/index.tsx
```diff
@@ -119,8 +119,8 @@ const AutocompleteField = ({
             'aria-label': label,
           }}
           required={required}
-          error={labelError}
-          helperText={labelError}
+          error={error}
```
bdauria, line 122:
```suggestion
          error={!isNil(error)}
```

Thebarda: We are using TextField from @centreon/uii that requires `string | undefined`

bdauria: In this case, I think helperText is redundant

src/InputField/Select/Autocomplete/index.tsx
```diff
@@ -119,8 +119,8 @@ const AutocompleteField = ({
...
+          helperText={error}
```
bdauria, line 123: I think helperText is already set when error is not nil within TextField

```suggestion
```

bdauria (approved):

jdelpierre (approved):

PR #293, opened 2020-12-03
chore(npm): Update frontend-core and fix ui-context to older version

bdauria:


package.json
```diff
@@ -31,7 +31,7 @@
     "react-dom": "^16.13.1"
   },
   "dependencies": {
-    "@centreon/ui-context": "github:centreon/ui-context",
+    "@centreon/ui-context": "github:centreon/ui-context#e085f8121536f3cac409cabb9eb847f211c4c5a5",
```
bdauria, line 34: We need to fix to an older version because the latest changes are not yet synchronized with centreon

Thebarda (approved):

jdelpierre (approved):

PR #296, opened 2020-12-11
fix(Autocomplete/Draggable): Fix z index

Thebarda:
This fixes a high z index to avoid that the dragged element disappear under MUI Dialog


jdelpierre (dismissed):

jdelpierre (approved):
Good idea to remove forced value :)

bdauria (approved):

PR #300, opened 2021-01-04
feat(Panel): Implement resizable panel

bdauria:
This allows the panel width to be resized:

resizable


Thebarda:
Maybe, add a small block wih an icon on the left side of the panel might be a good idea. It tells directly to the user that the panel is resizable

Thebarda (approved):

bdauria:
> Maybe, add a small block wih an icon on the left side of the panel might be a good idea. It tells directly to the user that the panel is resizable

What I see most of the time when areas are resizable, is that the only indicator is the 'resize' cursor while hovering the border. wdyt?

Thebarda:
> > Maybe, add a small block wih an icon on the left side of the panel might be a good idea. It tells directly to the user that the panel is resizable
> 
> What I see most of the time when areas are resizable, is that the only indicator is the 'resize' cursor while hovering the border. wdyt?

The simple cursor seems OK to me by the way :)

src/Panel/index.tsx
```diff
@@ -22,6 +23,8 @@ const useStyles = makeStyles<Theme, StylesProps>((theme) => ({
     height: '100%',
     display: 'grid',
     gridTemplate: 'auto auto 1fr / 1fr',
+    overflow: 'hidden',
+    width: ({ width }) => width,
```
jdelpierre, line 27: Why this syntax ?

bdauria: The width comes from the props, which are passed to the useStyles function. We can use it to create dynamic styles like we do here, to fix the width to the one coming from the props.

jdelpierre: Oh yes, I didn't pay attention to the useStyles() line (already present) which took "width" as a parameter.

jdelpierre (approved):

PR #311, opened 2021-01-13
fix(useLocaleDateTimeFormat): Use time format according to locale

bdauria:


Thebarda (approved):

jdelpierre (approved):
Why to change with AM/PM format ?

PR #324, opened 2021-02-23
enh(Listing): Use react window on listing

Thebarda:
This implements optimization to render a large number of rows without blow up the memory comsumption.
It uses `react-window` to render the list and `react-virtualized-auto-sizer` to handle height and width of the available space.


src/Listing/index.tsx
```diff
@@ -44,36 +44,42 @@ const useStyles = makeStyles<Theme>((theme) => ({
...
+    display: 'grid',
+    backgroundColor: theme.palette.common.white,
   },
 }));
 
+const rowHeight = 29;
+
+const itemKey = (index, data) => data.items[index].id;
+
+const createItemsData = memoize(({ items, properties }) => ({
```
kduret, line 78: since default equality function of memoize-one is `===`, useMemo does the same job ?

Thebarda: useMemo must be used inside a hook or a Component. memoize-one allow to use memoization outside them

Thebarda: But I think we can rid of it

kduret: Yes cause `createItemsData` is called inside Listing component

Thebarda: After some investigations, memoize-one is actually needed to avoid log-spike while scrolling / hovering elements

kduret (dismissed):

karimsx (approved):

src/Listing/index.tsx
```diff
@@ -269,20 +300,33 @@ const Listing = ({
...
         </div>
-        <Paper
+        <Table
+          size="small"
+          stickyHeader
+          className={classes.table}
           style={{
             maxHeight: tableMaxHeight(),
+            width: '100%',
+            height: '100%',
```
bdauria, line 310:
```suggestion
```

This is already declared in the table class

src/Listing/index.tsx
```diff
@@ -269,20 +300,33 @@ const Listing = ({
...
         >
-          <Table
-            size="small"
-            stickyHeader
-            className={classes.table}
+          <div
+            className={classes.listingHeader}
             style={{
               gridTemplateColumns: getGridTemplateColumn(),
+              paddingRight: subtract(
```
bdauria, line 318: Maybe extract this to a variable above

bdauria:
It seems to work properly after testing. 

One small warning that should be fixed though: 

image

Thebarda:
> One small warning that should be fixed though:

Fixed with the lastest commit

bdauria (dismissed):

bdauria:
After testing the latest changes against BAM, I found two new issues:

- It seems that the grid does not calculate the column template globally:

Capture d’écran 2021-02-26 à 16 47 49

- When the selected only items toggle changes, the height doesn’t switch back to the initial size afterwards:

Thebarda:
@bdauria For the second one, we should adapt some CSS in the secondary panel

Thebarda:
As decided in the frontend weekly meet, we close the PR because react-window adds to much constraint about CSS grid system that define our column construction

PR #329, opened 2021-03-03
enh(Listing): Add composite key and fix row color rendering

bdauria:
The composite key is necessary as it can happen that multiple entities have the same id (in which case the getCompositeId function must return a unique one)

### DIFF
--- a/src/Listing/index.tsx
+++ b/src/Listing/index.tsx
@@ -1,6 +1,6 @@
 import React, { useState, useRef, RefObject } from 'react';
 
-import { isNil } from 'ramda';
+import { equals, isNil } from 'ramda';
 
 import { makeStyles, Theme } from '@material-ui/core/styles';
 import {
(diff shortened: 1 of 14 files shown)


src/Listing/index.tsx
```diff
@@ -127,11 +126,10 @@ const Listing = ({
   onSort = (): void => undefined,
   labelDisplayedRows = ({ from, to, count }): string =>
     `${from}-${to} of ${count}`,
+  getCompositeId = ({ id }) => [id, id],
```
Thebarda, line 129:
```suggestion
  getCompositeId = ({ id }) => [id],
```
Maybe the default value should be a simple id

Thebarda (approved):

karimsx (approved):

Jordan91 (approved):

jdelpierre (approved):

PR #339, opened 2021-03-24
enh(TextField): Add a small variant on TextField

Thebarda:


karimsx (approved):

jdelpierre (approved):

src/InputField/Text/index.tsx
```diff
@@ -66,6 +71,7 @@ const TextField = React.forwardRef(
       ariaLabel,
       transparent = false,
       compact = false,
+      small = false,
```
kduret, line 74: it's a bit confusing with `compact` property, because both update fontSize
Is it possible to merge `compact` and `small` in a single property ? (which could be an enum)

Thebarda: Yep, like the `variant` in MUI

bdauria: Another interesting question: do we still need the compact version or can we do without it?

kduret (dismissed):

karimsx (dismissed):

bdauria:
@Thebarda After looking at the generated storyshots, it looks like you accidentally made the Input field bigger by default

Thebarda:
> @Thebarda After looking at the generated storyshots, it looks like you accidentally made the Input field bigger by default

Fixed

bdauria (approved):

kduret (approved):

PR #340, opened 2021-03-25
[Breaking] feat(Listing): Add possibliity to sort and change column display

bdauria:
This ads two features to the Listing:

- Column sorting
- Column display edit

This is done by passing a new prop `columnConfiguration` prop.

The Listing has been cleaned up a little and there is a breaking change at the prop naming level. 

https://www.loom.com/share/2c454eb0494f4b1199e110f92d073694


src/InputField/Select/Autocomplete/Multi/IconPopover/index.tsx
```diff
@@ -0,0 +1,77 @@
...
+    setAnchorEl(event.currentTarget);
+  };
+
+  return (
+    <ClickAwayListener onClickAway={close}>
+      <div>
+        <IconButton title={title} ariaLabel={title} onClick={toggle}>
+          {icon}
+        </IconButton>
+        <Popper
```
Thebarda, line 53: The `Popover` might be better here because it handle `ClickAwayListener`
https://material-ui.com/components/popover/#popover

Thebarda (commented):
Very well done 👍 
I think it is better to use the following icon as a drag handle
dragIndicator

Thebarda (approved):

src/Listing/Header/SortableCell/Content.tsx
```diff
@@ -0,0 +1,80 @@
...
+    };
+
+    return (
+      <div className={classes.content} ref={ref}>
+        {columnConfiguration?.sortable && (
+          <div className={classes.dragHandle} {...props}>
+            <DragIndicatorIcon fontSize="small" />
+          </div>
+        )}
+        {column.sortable === false ? (
```
Thebarda, line 63:
```suggestion
        {not(column.sortable) ? (
```

src/Listing/Header/SortableCell/Content.tsx
```diff
@@ -0,0 +1,80 @@
...
+      ...props
+    }: Props,
+    ref: React.ForwardedRef<HTMLDivElement>,
+  ): JSX.Element => {
+    const classes = useStyles({ isDragging });
+
+    const columnSortField = column.sortField || column.id;
+
+    const sort = (): void => {
+      const isDesc = columnSortField === sortField && sortOrder === 'desc';
```
Thebarda, line 48:
```suggestion
      const isDesc = and(equals(columnSortField, sortField), equals(sortOrder, 'desc'));
```

src/Listing/Header/SortableCell/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+    isSorting,
+    transform,
+  });
+  const cellClasses = useCellStyles({ listingCheckable: true });
+
+  return (
+    <HeaderCell
+      key={column.id}
+      padding={column.compact ? 'none' : 'default'}
+      sortDirection={sortField === column.id ? sortOrder : false}
```
Thebarda, line 72:
```suggestion
      sortDirection={equals(sortField, column.id) ? sortOrder : false}
```

src/Listing/index.tsx
```diff
@@ -72,64 +85,66 @@ const useStyles = makeStyles<Theme>((theme) => ({
...
   loading?: boolean;
   loadingDataMessage?: string;
   paginated?: boolean;
-  selectedRows?;
-  sorto?: 'asc' | 'desc';
-  sortf?: string;
-  tableData?;
-  totalRows?;
+  selectedRows?: Array<TRow>;
+  sortOrder?: 'asc' | 'desc';
```
Thebarda, line 101: Extract `asc` and `desc` into an enumeration to avoid value de-duplication

Thebarda (approved):

kduret (approved):

PR #350, opened 2021-04-14
fix(Autocomplete): Fix equality checking and prevent multi-select close when options are selected

bdauria:
autotomplete

### DIFF
--- a/src/InputField/Select/Autocomplete/index.tsx
+++ b/src/InputField/Select/Autocomplete/index.tsx
@@ -106,7 +106,7 @@ const AutocompleteField = ({
 }: Props): JSX.Element => {
   const classes = useStyles();
 
-  const areSelectEntriesEqual = (a, b) => {
+  const areSelectEntriesEqual = (option, value) => {
     const identifyingProps = ['id', 'name'];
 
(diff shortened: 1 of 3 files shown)


src/InputField/Select/Autocomplete/index.tsx
```diff
@@ -106,14 +106,20 @@ const AutocompleteField = ({
 }: Props): JSX.Element => {
   const classes = useStyles();
 
+  const areSelectEntriesEqual = (a, b) => {
```
Thebarda, line 109:
```suggestion
  const areSelectEntriesEqual = (option, value) => {
```

src/InputField/Select/Autocomplete/index.tsx
```diff
@@ -106,14 +106,20 @@ const AutocompleteField = ({
...
+    const identifyingProps = ['id', 'name'];
+
+    return equals(pick(identifyingProps, a), pick(identifyingProps, b));
```
Thebarda, line 112:
```suggestion
    return equals(pick(identifyingProps, option), pick(identifyingProps, value));
```

Thebarda (approved):

jdelpierre (approved):

PR #352, opened 2021-04-21
Fix(Listing): Reduce Listing freeze

Thebarda:
This reduces listing freeze by improving the DataCell memoization and not re-render the rows which are not in the viewport.

Before: For 100 elements on Resource Status, it took ~650ms  to render the Listing component
Now: It take ~30ms to render the Listing component

### DIFF
--- a/src/Listing/Row.tsx
+++ b/src/Listing/Row.tsx
@@ -2,7 +2,7 @@
 
 import * as React from 'react';
 
-import { equals, length, not, pluck } from 'ramda';
+import { equals, not, pluck } from 'ramda';
 
 import { TableRowProps, TableRow, makeStyles, Theme } from '@material-ui/core';
 
--- a/src/Listing/Row.tsx
+++ b/src/Listing/Row.tsx
@@ -2,17 +2,29 @@
 
 import * as React from 'react';
 
-import { equals } from 'ramda';
+import { equals, not } from 'ramda';
 
 import { TableRowProps, TableRow, makeStyles, Theme } from '@material-ui/core';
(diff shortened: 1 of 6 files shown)


src/Listing/Row.tsx
```diff
@@ -80,4 +107,20 @@ const Row = React.memo<Props>(
...
 );
 
-export default Row;
+const IntersectionRow = (props: Props): JSX.Element => {
+  const rowRef = React.useRef<HTMLDivElement | null>(null);
+  const { isInViewport, setElement } = useViewportIntersection();
+  const classes = useStyles();
+
+  React.useEffect(() => {
+    setElement(rowRef.current?.firstChild?.firstChild as HTMLDivElement);
```
bdauria, line 116: what does the first child of the first child represent herer? (meybe create an explicitely named function for this)

src/utils/useViewportIntersection.ts
```diff
@@ -0,0 +1,38 @@
+import * as React from 'react';
+
+interface GraphIntersectionState {
```
bdauria, line 3: I dont think that Graph means anything in this context

src/Listing/Row.tsx
Thebarda: It represent the first cell element (the checkbox or another component). Why do I not use rowRef.current ? Because of the `display: contents`. Use display contents on a div make it as an element (like block or inline) so the intersection observer will not work. That's why I need the closest element that is not a display contents.
But yes, I'm gonna make a function to retrieve this child

bdauria (approved):

jdelpierre (approved):

